PR #26, opened 2020-08-21
Develop

Andrii256:
https://andrii256.github.io/js_task_parse_list_DOM/


src/scripts/main.js
```diff
@@ -1,3 +1,15 @@
 'use strict';
 
-// write code here
+function sortList(list, fatherSelector) {
+  list.sort((a, b) => b.dataset.salary.slice(1).split(',').join('')
+    - a.dataset.salary.slice(1).split(',').join(''));
```
aliceblunt8, line 5: I think it's better to write a separate helper function for converting a and b to number

src/scripts/main.js
```diff
@@ -1,3 +1,15 @@
...
+
+  for (const person of list) {
+    document.querySelector(fatherSelector).append(person);
+  }
+}
+
+const listSelector = 'ul';
+const people = [...document.querySelector(listSelector).children];
+
+sortList(people, listSelector);
```
aliceblunt8, line 15: and where's getEmployees function?

Andrii256: I think it was only recomendations, but I found better way

src/scripts/main.js
Andrii256: i think here it is unnecessary. Only 2 ros of code

irenhh (commented):
you need to create two functions according to the task

Andrii256:
@irenhh я коли перший раз запустив деплой - закрив зарання термінал. зараз глючить. Не деплоїться https://prnt.sc/u63h1x

Але так то все працює, можеш запустити в себе

irenhh (approved):

PR #45, opened 2020-08-24
completed

DmJSCourse:
https://dmjscourse.github.io/js_task_parse_list_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,39 @@
 'use strict';
 
-// write code here
+const employees = document.getElementsByTagName('li');
+
+function sortList(list) {
+  const array = [...list];
+
+  array.sort((first, second) =>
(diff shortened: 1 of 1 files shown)


VenomSRT (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
...
+
+  checkingArray = checkingArray.filter(element => !isNaN(element));
+
+  return Number(checkingArray.join(''));
+}
+
+function getEmployees(list) {
+  const result = [];
+
+  [...list].forEach(element =>
```
irenhh, line 26: there is no need create a copy of the array

irenhh (approved):

src/scripts/main.js
DmJSCourse: it is html collection, not an array, and does not have forEach method

PR #127, opened 2020-12-30
Develop

xipholena:


Serhii-Alieksieievych (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,46 @@
...
+      salary: +(list[person].dataset.salary).slice(1).replace(/,/g, ''),
+      age: list[person].dataset.age,
+    });
+  }
+
+  const sortedBySalary = [ ...people ].sort((a, b) => +b.salary - +a.salary);
+
+  return sortedBySalary;
+}
+sortList(parsed);
```
varseniuk, line 21: You repeating yourself two times. 
In function `sortList` there's no need in any other fields except the 'salary'. You just need to get users' salaries from the list and sort them and then append them to the `list` element. Let's change your 'parsed' variable's value to `document.querySelector('ul')`. After that, no need to create a new array `people` and you can get the array of employees with [...list.children] inside your function and use some array methods like 'sort', 'map' and 'forEach'.

src/scripts/main.js
```diff
@@ -1,3 +1,46 @@
...
+    people.push({
+      name: list[person].innerText,
+      position: list[person].dataset.position,
+      salary: list[person].dataset.salary,
+      age: list[person].dataset.age,
+    });
+  }
+
+  return people;
+}
```
varseniuk, line 36: If you use `document.querySelector('ul')` you can get the employees list and map it to get the new array of objects without `for` loop.

src/scripts/main.js
```diff
@@ -1,3 +1,46 @@
...
+getEmployees(parsed);
+
+const sortedWithFunction = sortList(parsed);
+
+for (const coworker in parsed) { // for (const coworker in list)
+  parsed[coworker].innerText = sortedWithFunction[coworker].name;
+  parsed[coworker].dataset.position = sortedWithFunction[coworker].position;
+  parsed[coworker].dataset.salary = sortedWithFunction[coworker].salary;
+  parsed[coworker].dataset.age = sortedWithFunction[coworker].age;
+}
```
varseniuk, line 46: If you will sort your employees and append them right in the function `sortList` there will be no need in this part of the code.

varseniuk (changes requested):
DRY. Let's make your code much shorter and simplify it

varseniuk (approved):
Great work!

PR #128, opened 2020-12-30
solution

dtsehelnyk:
[DEMO LINK](https://dtsehelnyk.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,29 @@
 'use strict';
 
-// write code here
+const employeesBox = document.querySelector('ul');
+const employees = document.querySelectorAll('li');
+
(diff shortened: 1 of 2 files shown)


Serhii-Alieksieievych (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
 'use strict';
 
-// write code here
+const employeesBox = document.querySelector('ul');
+const employees = document.querySelectorAll('li');
```
varseniuk, line 4: You could only create one variable 'employeesBox' and use it like
`[...employeesBox.children]` instead of creating another variable 'employees'

varseniuk (approved):

PR #131, opened 2021-01-01
solution

khrulenko:
- [DEMO LINK](https://khrulenko.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,44 @@
 'use strict';
 
-// write code here
+const salaryToNumber = function(salary) {
+  return +salary.slice(1).split(',').join('');
+};
+
+const sortList = function(anylist) {
+  return anylist.sort(
+    (prev, next) =>
+      salaryToNumber(next.dataset.salary)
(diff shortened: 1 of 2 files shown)


olesiatk (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
...
+
+const sortList = function(anylist) {
+  return anylist.sort(
+    (prev, next) =>
+      salaryToNumber(next.dataset.salary)
+      - salaryToNumber(prev.dataset.salary)
+  );
+};
+
+const getEmployees = function(someList) {
```
vsm2k17, line 15: you can use the arrow function and map method

src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
...
+  const workers = [];
```
vsm2k17, line 16: you can return the result of the map method, without extra variables

vsm2k17 (approved):
good job, but you can make your code shorter

PR #132, opened 2021-01-03
Sorted list

vorfolomeevav:


5Mountains (approved):
Nice solution! 👍

src/scripts/main.js
```diff
@@ -1,3 +1,25 @@
 'use strict';
 
-// write code here
+const employeesCollection = document.querySelectorAll('ul > li');
+const employeesCollectionRepository = document.querySelector('ul');
```
varseniuk, line 4: You could create only one variable 'employeesCollectionRepository' (strange name) and use it like `[...employeesCollectionRepository.children]` instead of creating another variable 'employeesCollection'

src/scripts/main.js
```diff
@@ -1,3 +1,25 @@
...
+
+function sortList(collection, list) {
+  const sortedList = [...collection].sort((a, b) => {
+    return Number(b.dataset.salary.replace(/\D/g, ''))
+      - Number(a.dataset.salary.replace(/\D/g, ''));
```
varseniuk, line 9: It would be better to create a helper function for converting. Just FYI

varseniuk (approved):
Good. Left some small remarks

PR #139, opened 2021-01-04
solution

Ruslanwp:
- [DEMO LINK](https://Ruslanwp.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,21 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('ul');
+const persons = document.querySelectorAll('ul li');
+
+const sorted = [...persons].sort((a, b) => {
+  return Number(b.dataset.salary
+    .slice(1)
+    .split(',')
+    .join(''))
+  - Number(a.dataset.salary
+    .slice(1)
+    .split(',')
+    .join(''));
+});
+
+list.innerHTML = '';
+
+for (const li of sorted) {
+  list.appendChild(li);
+};


MarchD (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,21 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('ul');
+const persons = document.querySelectorAll('ul li');
```
varseniuk, line 4: You could create only one variable 'list' and then use it like `[...list.children]` to get the array of employees instead of creating the variable 'persons'

src/scripts/main.js
```diff
@@ -1,3 +1,21 @@
...
+
+const sorted = [...persons].sort((a, b) => {
+  return Number(b.dataset.salary
+    .slice(1)
+    .split(',')
+    .join(''))
+  - Number(a.dataset.salary
+    .slice(1)
+    .split(',')
+    .join(''));
```
varseniuk, line 14: According to the terms of the task, you should create a separate function for converting the numbers.

src/scripts/main.js
```diff
@@ -1,3 +1,21 @@
...
+});
+
+list.innerHTML = '';
+
+for (const li of sorted) {
+  list.appendChild(li);
+};
```
varseniuk, line 21: after `sort` method you can use `forEach` method and no need in this last part of the code.
E.g.
`[...persons].sort((a,b) => ...).forEach(person => list.append(person))`

varseniuk (changes requested):
> Your task: Sort list by salary. Get an array of employees. Write two functions:
- first, which sorts the list by salary from data attributes
- second, which returns an array of objects, where objects are employees.

Where is your second function? 

> 3. Create function getEmployees, which takes a list property, and returns an array of right objects.

>

varseniuk (approved):
Good

PR #141, opened 2021-01-04
solution

maryna-zinchenko:
[DEMO LINK](https://maryna-zinchenko.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,33 @@
 'use strict';
 
-// write code here
+const employees = document.querySelectorAll('li');
+const empList = document.querySelector('ul');
+
+function sortList(list) {
+  const sorted = [...list].sort((a, b) => toNum(b.dataset.salary)
+  - toNum(a.dataset.salary));
+
+  empList.append(...sorted);
+};
+
+function toNum(string) {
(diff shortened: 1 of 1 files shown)


zhenya-mezhueva (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,33 @@
 'use strict';
 
-// write code here
+const employees = document.querySelectorAll('li');
+const empList = document.querySelector('ul');
```
varseniuk, line 4: You could create only one variable 'empList' (strange name) and use it like [... empList.children] instead of creating another variable 'employees'

src/scripts/main.js
```diff
@@ -1,3 +1,33 @@
...
+    result.push({
+      name: list[i].innerText,
+      position: list[i].dataset.position,
+      salary: list[i].dataset.salary,
+      age: list[i].dataset.age,
+    });
+  }
+
+  return result;
+}
```
varseniuk, line 30: You cold just return `[...list.children].map(...)` instead of creating 'result' variable and 'for' loop

varseniuk (approved):
Nice, but there's some small remarks

PR #148, opened 2021-01-04
Solution

Dima-Semenov:
[DEMO LINK](https://Dima-Semenov.github.io/js_task_parse_list_DOM/)


5Mountains (approved):
Great solution! 🥇

varseniuk (approved):
You could create only 'list' variable and use it like `[...list.children]` instead of creating 'arrayLi' variable

PR #154, opened 2021-01-04
sort list by salary & get an array of objects with employee data

KirillLutsenko:
[DEMO LINK](https://KirillLutsenko.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,32 @@
 'use strict';
 
-// write code here
+const employees = [...document.querySelectorAll('li')];
+const employeesList = document.querySelector('ul');
+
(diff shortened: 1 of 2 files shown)


zhenya-mezhueva (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
 'use strict';
 
-// write code here
+const employees = [...document.querySelectorAll('li')];
+const employeesList = document.querySelector('ul');
```
varseniuk, line 4: You could create only one variable 'employeesList' and use it like [... employeesList.children] instead of creating another variable 'employees'

varseniuk (approved):

PR #155, opened 2021-01-04
solution

NazarDovhanchuk:
[DEMO LINK](https://nazardovhanchuk.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,35 @@
 'use strict';
 
-// write code here
+const employees = document.querySelectorAll('li');
+const empList = document.querySelector('ul');
+
+function convertToNumber(string) {
+  return +string.replace(/\D/g, '');
+};
+
+function sortList(list) {
+  const sorted = [...list].sort((a, b) => convertToNumber(b.dataset.salary)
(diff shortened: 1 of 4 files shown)


5Mountains (approved):
Awesome work! 🥇

src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
 'use strict';
 
-// write code here
+const employees = document.querySelectorAll('li');
+const empList = document.querySelector('ul');
```
varseniuk, line 4: You could create only one variable 'empList' and use it like [... empList.children] instead of creating another variable 'employees'

src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
...
+      position: person.dataset.position,
+      salary: person.dataset.salary,
+      age: person.dataset.age,
+    };
+
+    result.push(personInformation);
+  }
+
+  return result;
+}
```
varseniuk, line 32: You cold just return [...list.children].map(...) instead of creating 'result' variable and 'for' loop

varseniuk (approved):

PR #286, opened 2021-08-12
solution

obordiug:
https://obordiug.github.io/js_task_parse_list_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,43 @@
 'use strict';
 
-// write code here
+const listOfEmployees = document.querySelector('ul');
+
+function conversionToNumber(string) {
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,43 @@
...
+
+    const dataElement = {
+      ...elementName,
+      ...element.dataset,
+    };
+
+    dataAllEmployees.push(dataElement);
+  });
+
+  return dataAllEmployees;
```
vkryvytskyy, line 40:
```suggestion
   return [ ...list.children ].map(element => {
    const elementName = { name: element.innerText };

    return dataElement = {
      ...elementName,
      ...element.dataset,
    };
  });
```

vkryvytskyy (approved):

PR #290, opened 2021-08-14
employees sorted

ihor-jpeg:
- [DEMO LINK](https://ihor-jpeg.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_task_parse_list_DOM/)
+    - [DEMO LINK](https://ihor-jpeg.github.io/js_task_parse_list_DOM/)
 2. Follow [this instructions](https://mate-academy.github.io/layout_task-guideline/)
     - Run `npm run test` command to test your code;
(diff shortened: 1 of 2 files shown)


Anna-Tkachenko (changes requested):
Good Job👍

But the task was:
Create function sortList, which takes a list property, and sorts list items in descending order by salary from data attribute. (Salary from data attribute is a string, do not forget to convert it to number, write helper function for it). Append sorted items to list.

Create function getEmployees, which takes a list property, and returns an array of right objects.

Call both functions.

Your function getEmployeesObjects is right, but sortBySalary function should work as in task description.

ihor-jpeg:
> Good Job👍
> 
> But the task was:
> Create function sortList, which takes a list property, and sorts list items in descending order by salary from data attribute. (Salary from data attribute is a string, do not forget to convert it to number, write helper function for it). Append sorted items to list.
> 
> 
> Call both functions.
> 
> Your function getEmployeesObjects is right, but sortBySalary function should work as in task description.

Thank a great point. i didn't pay attention. but then i do not really understand why we need the function for getting objects. it would make sense to use those objects for appending. strange..

PR #312, opened 2021-09-25
solution

Bohdan-Kochetov:


src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
 'use strict';
 
-// write code here
+const list = document.querySelectorAll('li');
```
Anna-Tkachenko, line 3: If you get all 'li' elements it's listItems, not list
```suggestion
const listItems = document.querySelectorAll('li');
```

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+      name: item.innerText,
+      ...item.dataset,
+    }));
+
+  people.sort((a, b) => convert(b.salary) - convert(a.salary));
+
+  function convert(str) {
+    return +str.replace(/\D/g, '');
+  }
+}
```
Anna-Tkachenko, line 18: Function sortList should sort the list by salary from data attributes, not from objects properties
image

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #334, opened 2021-11-19
Solution

dariia-romanova:
[DEMO LINK](https://dariia-romanova.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,50 @@
 'use strict';
 
-// write code here
+function convertToNumber(str) {
+  return str.split('').filter(element => {
+    if (isNaN(+element)) {
+      return false;
+    }
+
+    return true;
+  }).join('');
+}
+
+const listOfEmployees = document.querySelector('ul');
+
+function sortList(list) {
+  const children = list.children;
+  const employees = [...children];
+
+  employees.sort((employee1, employee2) => {
+    const salary1 = employee1.dataset.salary;
+    const salary2 = employee2.dataset.salary;
+
+    return convertToNumber(salary2) - convertToNumber(salary1);
+  });
+
+  list.innerHTML = `
+  ${employees.map(employee => (`
+    <li
+      data-position = "${employee.dataset.position}"
+      data-salary = "${employee.dataset.salary}"
+      data-age = "${employee.dataset.age}"
+    >${employee.innerHTML.trim()}
+    </li>`
+  )).join('')}`;
+}
+
+function getEmployees(list) {
+  const children = list.children;
+  const employees = [...children];
+
+  return employees.map((employee) => ({
+    name: employee.innerHTML.trim(),
+    position: employee.dataset.position,
+    salary: employee.dataset.salary,
+    age: employee.dataset.age,
+  }));
+}
+
+sortList(listOfEmployees);
+getEmployees(listOfEmployees);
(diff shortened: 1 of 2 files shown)


SerhiiKirik (approved):
Good approach 🔥

src/scripts/main.js
```diff
@@ -1,3 +1,50 @@
...
+  list.innerHTML = `
+  ${employees.map(employee => (`
+    <li
+      data-position = "${employee.dataset.position}"
+      data-salary = "${employee.dataset.salary}"
+      data-age = "${employee.dataset.age}"
+    >${employee.innerHTML.trim()}
+    </li>`
+  )).join('')}`;
+}
```
AlinaMatuschak, line 35: this function name is sortList. But there you also change dom. Maybe better rename function or change dom outside on the bottom of the file?

src/scripts/main.js
```diff
@@ -1,3 +1,50 @@
...
+  const employees = [...children];
+
+  employees.sort((employee1, employee2) => {
+    const salary1 = employee1.dataset.salary;
+    const salary2 = employee2.dataset.salary;
+
+    return convertToNumber(salary2) - convertToNumber(salary1);
+  });
+
+  list.innerHTML = `
```
AlinaMatuschak, line 26: try use insertAdjacentHTML

src/scripts/main.js
```diff
@@ -1,3 +1,50 @@
...
+function getEmployees(list) {
+  const children = list.children;
+  const employees = [...children];
+
+  return employees.map((employee) => ({
+    name: employee.innerHTML.trim(),
+    position: employee.dataset.position,
+    salary: employee.dataset.salary,
+    age: employee.dataset.age,
+  }));
```
AlinaMatuschak, line 46:
```suggestion
  return employees.map((employee) => ({
    name: employee.innerHTML.trim(),
    ...person.dataset,
  }));
```

src/scripts/main.js
```diff
@@ -1,3 +1,50 @@
...
+
+    return true;
+  }).join('');
+}
+
+const listOfEmployees = document.querySelector('ul');
+
+function sortList(list) {
+  const children = list.children;
+  const employees = [...children];
```
AlinaMatuschak, line 17:
```suggestion
  const employees = [...list.children];
```

src/scripts/main.js
```diff
@@ -1,3 +1,50 @@
...
+  return employees.map((employee) => ({
+    name: employee.innerHTML.trim(),
+    position: employee.dataset.position,
+    salary: employee.dataset.salary,
+    age: employee.dataset.age,
+  }));
+}
+
+sortList(listOfEmployees);
+getEmployees(listOfEmployees);
```
AlinaMatuschak, line 50: this function returns employees with names. But you don't use this value 🤔

AlinaMatuschak (changes requested):
try to use employees obj with names to update dom

src/scripts/main.js
dariia-romanova: But I need to replace html rather than add it. Can't I use innerHTML in this case?

AlinaMatuschak: You can leave default list

AlinaMatuschak (approved):
🔥

PR #335, opened 2021-11-22
Solution

SerhiiKirik:
- [DEMO LINK](https://SerhiiKirik.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,34 @@
 'use strict';
 
-// write code here
+const listOfEmployees = document.querySelector('ul');
+
+function getEmployees(list) {
+  const employees = [...list.children];
+
+  return employees;
+}
+
+function sortList(list) {
+  const callBack = (el) => (
+    +el.dataset.salary.replace(/[$,]/g, '')
+  );
+
+  const salary = getEmployees(listOfEmployees).sort((a, b) => (
+    callBack(b) - callBack(a)
+  ));
+
+  list.innerHTML = `
+    ${salary.map(employee => `
+      <li
+        data-position="${employee.dataset.position}"
+        data-salary="${employee.dataset.salary}"
+        data-age="${employee.dataset.age}"
+      >
+        ${employee.innerText}
+      </li>
+    `).join('')}
+  `;
+}
+
+getEmployees(listOfEmployees);
+sortList(listOfEmployees);
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_task_parse_list_DOM/)
+    - [DEMO LINK](https://SerhiiKirik.github.io/js_task_parse_list_DOM/)
 2. Follow [this instructions](https://mate-academy.github.io/layout_task-guideline/)
     - Run `npm run test` command to test your code;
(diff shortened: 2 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
 'use strict';
 
-// write code here
+const listOfEmployees = document.querySelector('ul');
+
+function getEmployees(list) {
+  const employees = [...list.children];
+
+  return employees;
+}
```
dariia-romanova, line 9: This function returns array of li element. It should return array of objects that looks like this instead:

> second, which returns an array of objects, where objects are employees.
> The schema for the employee:
> 
> ```
> {
>   name, 
>   position,
>   salary,
>   age
> }
> ```

dariia-romanova (changes requested):

dariia-romanova (approved):

src/scripts/main.js
SerhiiKirik: Mb you right, but i think, that immediately add the finished result is not a problem)

src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+
+function sortList(list) {
+  const callBack = (el) => (
+    +el.dataset.salary.replace(/[$,]/g, '')
+  );
```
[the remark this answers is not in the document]
SerhiiKirik: My bad)

src/scripts/main.js
illiarerun: `3. Create function getEmployees, which takes a list property, and returns an array of right objects.`
So yeah, let's do it that way

src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+        data-age="${employee.dataset.age}"
+      >
+        ${employee.innerText}
+      </li>
+    `).join('')}
+  `;
+}
+
+getEmployees(listOfEmployees);
+sortList(listOfEmployees);
```
illiarerun, line 34: All you actually need is one call
```suggestion
sortList(listOfEmployees);
```

src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+
+  return employees;
+}
+
+function sortList(list) {
+  const convertSalary = (el) => (
+    +el.dataset.salary.replace(/[$,]/g, '')
+  );
+
+  const salary = getEmployees(listOfEmployees).sort((a, b) => (
```
illiarerun, line 16: you can pass list from the closure

illiarerun (changes requested):
Good job 🔥  Let's work on fixes now:

src/scripts/main.js
```diff
@@ -1,3 +1,41 @@
...
+  return employees.map((employee) => ({
+    name: employee.innerHTML.trim(),
+    position: employee.dataset.position,
+    salary: employee.dataset.salary,
+    age: employee.dataset.age,
+  }));
+}
+
+function sortList(list) {
+  const employees = [...list.children];
```
illiarerun, line 17: You need to use your getEmployees func here

illiarerun (changes requested):

src/scripts/main.js
SerhiiKirik: 4.Call both functions.

illiarerun (approved):
LGTM 🔥

PR #351, opened 2021-12-28
Develop

anna-tovkailo:


src/scripts/main.js
```diff
@@ -1,3 +1,65 @@
 'use strict';
 
-// write code here
+const queryList = document.querySelector('ul').children;
+
+const getEmployees = function(list) {
+  const arrayList = [...list].map(item => {
```
aliceblunt8, line 6:
```suggestion
  const arrayList = list.map(item => {
```

src/scripts/main.js
```diff
@@ -1,3 +1,65 @@
...
+      return 1;
+    }
+
+    return 0;
+  });
+
+  return list;
+};
+
+sortList(listForSort);
```
aliceblunt8, line 46:
```suggestion
const sortList = function(list) {
  list.sort(function(a, b) {
    let aSort = a.salary.slice(1, 10).split(',');

    aSort = Number(aSort.join(''));

    let bSort = b.salary.slice(1, 10).split(',');

    bSort = Number(bSort.join(''));

    if (aSort > bSort) {
      return -1;
    }

    if (aSort < bSort) {
      return 1;
    }

    return 0;
  });

  return list;
};

sortList(getEmployees(queryList));
```

src/scripts/main.js
```diff
@@ -1,3 +1,65 @@
...
+
+const sortList = function(list) {
+  list.sort(function(a, b) {
+    let aSort = a.salary.slice(1, 10).split(',');
+
+    aSort = Number(aSort.join(''));
+
+    let bSort = b.salary.slice(1, 10).split(',');
+
+    bSort = Number(bSort.join(''));
```
aliceblunt8, line 30: you can create a separate function for formatting values before sorting them

aliceblunt8: DRY

aliceblunt8: after that you can reduce your code inside sort to smth like this
```
(a, b) => toNumber(b.salary) - toNumber(a.salary)
```

src/scripts/main.js
```diff
@@ -1,3 +1,65 @@
...
-// write code here
+const queryList = document.querySelector('ul').children;
+
+const getEmployees = function(list) {
+  const arrayList = [...list].map(item => {
+    const itemEmployee = {
+      name: item.innerHTML.trim(),
+      position: item.getAttribute('data-position'),
+      salary: item.getAttribute('data-salary'),
+      age: item.getAttribute('data-age'),
```
aliceblunt8, line 11:
```suggestion
      ...item.dataset,
```

aliceblunt8: this should work too

aliceblunt8: then just position,  salary  and age will be stored inside dataset object

src/scripts/main.js
```diff
@@ -1,3 +1,65 @@
...
+for (let i = 0; i < listForSort.length; i++) {
+  const item = document.createElement('li');
+
+  item.innerHTML = listForSort[i].name;
+  item.setAttribute('data-position', listForSort[i].position);
+  item.setAttribute('data-salary', listForSort[i].salary);
+  item.setAttribute('data-age', listForSort[i].age);
+
+  thisList.append(item);
+}
```
aliceblunt8, line 65: why not just?

```
thisList.append(...listForSort);
```

aliceblunt8 (changes requested):

src/scripts/main.js
anna-tovkailo: I can't do that, list is a collection, there seems to be no way to use "map"

src/scripts/main.js
anna-tovkailo: It doesn't work either, the list is used below in a loop. Or there, too, must be changed to getEmployees?

src/scripts/main.js
anna-tovkailo: I moved the formatting into a separate function, but this syntax does not work for me

src/scripts/main.js
anna-tovkailo: fixed, thanks!

src/scripts/main.js
anna-tovkailo: listForSort is an array of objects, not "li", so I can't append it at once. And previous list cannot be deleted without "for" with "x.remove()"

varseniuk (approved):
👍

PR #390, opened 2022-04-02
Result

xpromt:


Bohdan-Maksymiuk (changes requested):
check demo link

xpromt:
https://xpromt.github.io/js_task_parse_list_DOM

src/scripts/main.js
```diff
@@ -1,3 +1,25 @@
...
+    const next = +y.dataset.salary.replace(',', '').replace('$', '');
+
+    return next - prev;
+  });
+
+  return empList.append(...list);
+}
+
+function getEmployees(list) {
+  return [...list].map(employee => ({
```
aliceblunt8, line 18: map method doesn't mutate the original list

```suggestion
  return list.map(employee => ({
```

aliceblunt8 (approved):

PR #405, opened 2022-04-18
develop

MariaBondar:


src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
...
+const employeesList = [...list.children];
+
+function salary(str) {
+  return +str.split('').filter(char => '0123456789'.includes(char)).join('');
+}
+
+function sortList(str) {
+  const sortedBySalary = str.sort((x, y) =>
+    salary(x.dataset.salary) - salary(y.dataset.salary)
+  );
```
ivan-kadykalo, line 13: You should make reverse list, with descent.

src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
...
+
+  sortedBySalary.map(element => list.append(element));
```
ivan-kadykalo, line 15: there is no need in map, you didn't return anything
forEach will work also

ivan-kadykalo (changes requested):

PR #424, opened 2022-06-01
solution

KatrukEvgeny:


src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
 'use strict';
 
-// write code here
+const list = document.querySelectorAll('li');
+
+function sortList() {
+  const arrPeople = [...list].map(el => {
```
aliceblunt8, line 6: map doesn't mutate array

```suggestion
  const arrPeople = list.map(el => {
```

src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
...
+      >
+        ${el['name']}
+      </li>`);
+  }).join('');
+}
+
+sortList(list);
+
+function getEmployees() {
+  return [...list].map(el => {
```
aliceblunt8, line 35: same

src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
...
+    return {
+      'name': el.innerText,
```
aliceblunt8, line 37: no need, under the hood it will be transformed to string anyway

```suggestion
      name: el.innerText,
```

src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
...
+      'position': el.dataset.position,
+      'salary': el.dataset.salary,
+      'salaryNum': +el.dataset.salary.replace(/[$,]/g, ''),
+      'age': el.dataset.age,
+    };
+  });
+
+  arrPeople.sort((a, b) => b.salaryNum - a.salaryNum);
```
aliceblunt8, line 16: I think it's unnecessary to create a new array just to add salaryNum property to it, use salary for sorting

src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
...
+  ulInsert.innerHTML = arrPeople.map(el => {
+    return (
+      `<li
+        data-position="${el['position']}"
+        data-salary="${el['salary']}"
+        data-age="${el['age']}"
+      >
+        ${el['name']}
+      </li>`);
+  }).join('');
```
aliceblunt8, line 29: magic 🪄 

```suggestion
 initialList.append(...sortedList);
```

src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
...
+      'position': el.dataset.position,
+      'salary': el.dataset.salary,
+      'salaryNum': +el.dataset.salary.replace(/[$,]/g, ''),
+      'age': el.dataset.age,
+    };
+  });
+
+  arrPeople.sort((a, b) => b.salaryNum - a.salaryNum);
+
+  const ulInsert = document.querySelector('ul');
```
aliceblunt8, line 18: also better to move out of the function, and place it near list variable at the beginning of the file

```suggestion
  const initialList = document.querySelector('ul');
```

aliceblunt8 (changes requested):

PR #459, opened 2022-07-06
Solution

fledith:
[DEMO LINK](https://fledith.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,34 @@
 'use strict';
 
-// write code here
+class Employer {
+  constructor(fullName, position, salary, age) {
+    this.name = fullName;
+    this.position = position;
+    this.salary = salary;
+    this.age = age;
+  }
+}
+
+const employers = [];
+
+const li = document.querySelectorAll('li');
+const fullNames = [...li].map(item => item.innerText);
+const positions = [...li].map(item => item.getAttribute('data-position'));
+const salaries = [...li].map(item =>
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+    this.salary = salary;
+    this.age = age;
+  }
+}
+
+const employers = [];
+
+const li = document.querySelectorAll('li');
+const fullNames = [...li].map(item => item.innerText);
+const positions = [...li].map(item => item.getAttribute('data-position'));
```
solaryasha, line 16: use dataset here

solaryasha (changes requested):
> Create function sortList, which takes a list property, and sorts list items in descending order by salary from data attribute. (Salary from data attribute is a string, do not forget to convert it to number, write helper function for it). Append sorted items to list.

Call both functions.

src/scripts/main.js
```diff
@@ -1,3 +1,47 @@
...
+    employer[employer.length] = new Employer(list[i].innerText,
+      list[i].dataset.position,
+      +list[i].dataset.salary.replace(/\D/g, ''),
+      list[i].dataset.age);
+  }
+
+  return employer;
+}
+
+function listSorted(list) {
```
solaryasha, line 27:
```suggestion
function sortList(list) {
```

solaryasha (approved):

PR #475, opened 2022-07-18
Solution

Lentyaeev:
[DEMO LINK](https://lentyaeev.github.io/js_task_parse_list_DOM/)


VETALEK (changes requested):
let's use array iteration methods instead of loops 😉

PR #476, opened 2022-07-18
Solution

TwoAndMore:
- [DEMO LINK](https://TwoAndMore.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,38 @@
 'use strict';
 
-// write code here
+const employeesList = [...document.querySelectorAll('li')];
+
+function sortList(list) {
+  list.sort((a, b) => b.salary - a.salary);
+}
+
+function getEmployees(list) {
+  const people = [];
+
+  for (const elem of list) {
+    const person = {};
+
+    person.name = elem.innerText;
+    person.position = elem.dataset.position;
+    person.salary = elem.dataset.salary.replace(/[$,]/g, '');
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,38 @@
 'use strict';
 
-// write code here
+const employeesList = [...document.querySelectorAll('li')];
+
+function sortList(list) {
+  list.sort((a, b) => b.salary - a.salary);
+}
```
VETALEK, line 7: either call it `sortBySalary` or remove this function and just create a callback with salary comparison 😉

src/scripts/main.js
```diff
@@ -1,3 +1,38 @@
...
+  }
+
+  return people;
+}
+
+const employees = getEmployees(employeesList);
+
+sortList(employees);
+
+for (let i = 0; i < employees.length; i++) {
```
VETALEK, line 30: array iteration methods 😿

VETALEK (changes requested):
Good start 🔥

src/scripts/main.js
TwoAndMore: ```Create function sortList, which takes a list property, and sorts list items in descending order by salary from data attribute. (Salary from data attribute is a string, do not forget to convert it to number, write helper function for it). Append sorted items to list.```

VETALEK (approved):
🚀

PR #486, opened 2022-07-22
ready

denenergi:
[DEMO LINK](https://denenergi.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,15 @@
 'use strict';
 
-// write code here
+const listItem = document.querySelector('.list');
+const item = document.querySelectorAll('.item');
+
+function sortList(list) {
+  return [...list].sort((a, b) =>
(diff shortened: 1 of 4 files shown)


fledith (changes requested):
You don't need to change styles or HTML layout in this task. Change only main.js file.

src/scripts/main.js
```diff
@@ -1,3 +1,15 @@
 'use strict';
 
-// write code here
+const listItem = document.querySelector('ul');
+const item = document.querySelectorAll('li');
+
+function sortList(list) {
+  return [...list].sort((a, b) =>
+    b.dataset.salary.replace(/\D/g, '') - a.dataset.salary.replace(/\D/g, ''));
```
Ihor-Karpyn, line 8:
```suggestion
  return [...list].sort((a, b) => (
    b.dataset.salary.replace(/\D/g, '') - a.dataset.salary.replace(/\D/g, '')
  ));
```

Ihor-Karpyn (approved):

PR #500, opened 2022-08-09
Solution

OksanaShvets1:
not sure is it correct

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,34 @@
 'use strict';
 
 // write code here
+
+const allLi = document.querySelectorAll('ul');
+
+function sortList(list) {
+  const listElements = Array.prototype.slice(list.children);
+  const sortedListElements = listElements.sort((a, b) =>
+    Number(b.getAttribute('data-salary').slice(1).replace(/,/g, ''))
+    - Number(a.getAttribute('data-salary').slice(1).replace(/,/g, '')));
+
+  // list.innerHTML = '';
+
+  sortedListElements.forEach(function(el) {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
 'use strict';
 
 // write code here
+
+const allLi = document.querySelectorAll('ul');
```
Bohdan-Maksymiuk, line 5:
```suggestion
const allLi = document.querySelector('ul');
```
there is one list

src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+  const sortedListElements = listElements.sort((a, b) =>
+    Number(b.getAttribute('data-salary').slice(1).replace(/,/g, ''))
+    - Number(a.getAttribute('data-salary').slice(1).replace(/,/g, '')));
+
+  // list.innerHTML = '';
+
+  sortedListElements.forEach(function(el) {
+    list.appendChild(el);
+  });
+}
```
Bohdan-Maksymiuk, line 18: try to sort in such way
```
const startList = document.querySelector('ul');

function format(element) {
  return +element.slice(1).split(',').join('');
}

function sortList(list) {
  const items = list.querySelectorAll('li');

  const arr = [...items].sort((person1, person2) => {
    return format(person2.dataset.salary) - format(person1.dataset.salary);
  });

  arr.forEach(person => list.append(person));
}
```

Bohdan-Maksymiuk (changes requested):

ArturErmolenko (approved):

PR #507, opened 2022-08-15
solution

AlexanderKolomiiets:
[DEMO LINK](https://alexanderkolomiiets.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -2,9 +2,13 @@
 
 const employees = [...document.querySelectorAll('li')];
 
+`function toNumber(string) {
+  return string.replace(/\D/g, '');
+}`
+
 function sortList(list) {
   return list.sort((a, b) => (
-    b.dataset.salary.replace(/\D/g, '') - a.dataset.salary.replace(/\D/g, '')));
+    toNumber(b.dataset.salary) - toNumber(a.dataset.salary)));
 }
 
 function getEmployees(list) {
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,31 @@
 'use strict';
 
-// write code here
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,31 @@
 'use strict';
 
-// write code here
+const employees = [...document.querySelectorAll('li')];
+
+function sortList(list) {
+  return list.sort((a, b) => (
+    b.dataset.salary.replace(/\D/g, '') - a.dataset.salary.replace(/\D/g, '')));
```
tiserett, line 7:
```suggestion
    b.dataset.salary.replace(/\D/g, '') - a.dataset.salary.replace(/\D/g, '')));
```

according to task, you should create function for converting salary string into number

src/scripts/main.js
```diff
@@ -1,3 +1,31 @@
...
+  list.forEach(element => {
+    result.push({
+      name: element.innerText,
+      position: element.dataset.position,
+      salary: element.dataset.salary,
+      age: element.dataset.age,
+    });
+  });
+
+  return result;
```
tiserett, line 24:
```suggestion
  const result = [];

  list.forEach(element => {
    result.push({
      name: element.innerText,
      position: element.dataset.position,
      salary: element.dataset.salary,
      age: element.dataset.age,
    });
  });

  return result;
```

You can get rid of variable result, just make copy of your list, which you take as parametr, and return the result of iteration through it with map function

tiserett (changes requested):
few small fixes required

tiserett (changes requested):
(Salary from data attribute is a string, do not forget to convert it to number, write helper function for it)

tiserett (approved):
gj

PR #540, opened 2022-09-13
Solution

AMahalias:
[DEMO LINK](https://AMahalias.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,39 @@
 'use strict';
 
-// write code here
+const fullList = document.querySelectorAll('li');
+
+function sortList(list) {
+  return [...list].sort((a, b) => {
+    const salaryA = +(a.dataset.salary.slice(1).split('')
+      .filter(item => item !== ',').join(''));
+    const salaryB = +(b.dataset.salary.slice(1).split('')
+      .filter(item => item !== ',').join(''));
+
+    return salaryB - salaryA;
+  });
+}
+
+const sortedList = sortList(fullList);
+
+document.querySelector('ul').innerHTML = sortedList.map(item => {
+  return item.outerHTML;
+}).join('');
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
...
+  [...list].map(element => {
+    const employee = {
+      name: element.innerText.trim(),
+      position: element.dataset.position,
+      salary: element.dataset.salary,
+      age: element.dataset.age,
+    };
+
+    employees.push(employee);
+  });
```
wa7er14, line 34: `map` returns new array, so you don't need to push items in `employees`, just return `employee` in `map` callback and return result of `map` in `getEmployees`

src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
 'use strict';
 
-// write code here
+const fullList = document.querySelectorAll('li');
+
+function sortList(list) {
+  return [...list].sort((a, b) => {
+    const salaryA = +(a.dataset.salary.slice(1).split('')
+      .filter(item => item !== ',').join(''));
```
wa7er14, line 8: better to use `replace` instead of `split.filter.join`. write helper function for parsing and use it for salary a/b

src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
...
+
+    return salaryB - salaryA;
+  });
+}
+
+const sortedList = sortList(fullList);
+
+document.querySelector('ul').innerHTML = sortedList.map(item => {
+  return item.outerHTML;
+}).join('');
```
wa7er14, line 20:
```suggestion
document.querySelector('ul').innerHTML = sortedList.map(item => item.outerHTML).join('');
```

wa7er14 (approved):

PR #544, opened 2022-09-15
Develop

JOWISSA:
[DEMO LINK](https://JOWISSA.github.io/js_task_parse_list_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,24 @@
 'use strict';
 
-// write code here
+
+const salaries = [...document.querySelectorAll('[data-salary')];
+
+function sortList(list) {
+  list.sort((a, b) =>
+    +b.dataset.salary.slice(1).split(',').join('')
```
wa7er14, line 8: move to helper function, use `replace`

src/scripts/main.js
```diff
@@ -1,3 +1,24 @@
...
+    - +a.dataset.salary.slice(1).split(',').join(''));
+
+  list.map(li => li.parentElement.append(li));
```
wa7er14, line 11: you chain `map` just after `sort`

wa7er14 (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,27 @@
 'use strict';
 
-// write code here
+const listUl = document.querySelector('ul');
+const listLi = [...document.querySelectorAll('[data-salary')];
+
+function sortList(list) {
+  const replace = (str) => {
```
ivanchuk-yurii, line 7: I was talking about `str.replace(',', '')`, not naming function replace)

ivanchuk-yurii (changes requested):

src/scripts/main.js
```diff
@@ -5,7 +5,7 @@ const listLi = [...document.querySelectorAll('[data-salary')];
 
 function sortList(list) {
   const replace = (str) => {
-    return +str.slice(1).split(',').join('');
```
ivanchuk-yurii: you should replace for empty string, not `'.'`
`138,575` - its 138 thousands and 575,  after replace `138.575`, it's just 138 and decimals, so your sorting isn't working correct

ivanchuk-yurii (changes requested):

src/scripts/main.js
SanyaBratashchuk, line 7:
```suggestion
  const normalizeNumber = (str) => {
```

SanyaBratashchuk (approved):

PR #546, opened 2022-09-18
Solution

MaxManis:


src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
 'use strict';
 
 // write code here
+const ulItem = document.querySelector('ul');
+const liItems = document.querySelectorAll('li');
+
+function sortList(list, listItems) {
+  const itemsSorted = [...listItems].sort((a, b) => (
+    +((b.dataset.salary)
```
OleksiiBlezniuk, line 9: Use `Number`, it's more readable then `+`

src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
...
+      .slice(1)
+      .split(',')
+      .join(''))
+    - +((a.dataset.salary)
+      .slice(1)
+      .split(',')
+      .join(''))
```
OleksiiBlezniuk, line 16: Will be more readable if you'll store result of this operations for both salaries to variables and return result of subtraction

OleksiiBlezniuk: Also would be nice to move this converting logic to helper function

src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
...
+  ));
+
+  list.innerHTML = '';
+
+  for (let i = 0; i < itemsSorted.length; i++) {
+    list.appendChild(itemsSorted[i]);
+  }
```
OleksiiBlezniuk, line 23: You can use `forEach` for this purpose

src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
...
+}
+
+function getEmployees(listItems) {
+  const result = [];
```
OleksiiBlezniuk, line 27: In this case you'll store eployees objects in this array, so better to specify that in name

OleksiiBlezniuk, line 27:
```suggestion
  const employees = [];
```

src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
...
+
+  for (const item of listItems) {
+    const employer = {
```
OleksiiBlezniuk, line 30:
```suggestion
    const employee = {
```

src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
...
+      name: item.innerText,
+      position: item.dataset.position,
+      salary: item.dataset.salary,
+      age: item.dataset.age,
+    };
+
+    result.push(employer);
+  }
```
OleksiiBlezniuk, line 38: You just modifying array items, why not using `map` method?

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (commented):
Also this task should be deployed

src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
...
+
+function getEmployees(listItems) {
+  const employees = [...listItems].map(item => (
+    {
+      name: item.innerText,
+      position: item.dataset.position,
+      salary: item.dataset.salary,
+      age: item.dataset.age,
+    }
+  ));
```
OleksiiBlezniuk, line 37:
```suggestion
  const employees = [...listItems].map(item => ({
      name: item.innerText,
      position: item.dataset.position,
      salary: item.dataset.salary,
      age: item.dataset.age,
  }));
```

OleksiiBlezniuk: It's ok, but usually we don't wrap curly brackets to new line

OleksiiBlezniuk: Also you could just return result of map here (without saving to variable)

OleksiiBlezniuk (approved):

PR #550, opened 2022-09-25
JS parse list DOM (first solution)

Oleksandr-Naumchak:
JS parse list DOM (first solution)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,28 @@
 'use strict';
 
-// write code here
+// const elements = document.body.querySelectorAll('li');
+const elements = [...document.querySelectorAll('li')];
+
+function convertToNumber(inputString) {
+  return +inputString.replace(/\D/g, '');
+}
+
+function sortList(list) {
+  const employeeList = document.querySelector('ul');
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
...
+
+  const sortedList = [...list].sort((prev, next) =>
+    convertToNumber(next.dataset.salary)
+    - convertToNumber(prev.dataset.salary));
+
+  employeeList.append(...sortedList);
+};
+
+function getEmployees(list) {
+  return [...list].map((employee) => ({
```
SanyaBratashchuk, line 21:
```suggestion
  return list.map((employee) => ({
```

SanyaBratashchuk (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
 'use strict';
 
-// write code here
+// const elements = document.body.querySelectorAll('li');
```
marina-tilniak, line 3: don't leave commented code

marina-tilniak (approved):

PR #558, opened 2022-09-30
solution was added

Maksym845:


src/scripts/main.js
```diff
@@ -1,3 +1,26 @@
 'use strict';
 
-// write code here
+const salaryConvertion = (string) => string.replace(/[^\d]/g, '');
+
+const sortList = (list) => {
+  const sortedList = list.sort((first, second) => {
```
Denchepornyuk, line 6: sort is mutating method, let's create a copy for list array

Denchepornyuk (approved):

PR #562, opened 2022-10-05
made parse list

Hlazman:
DEMO LINK
https://hlazman.github.io/js_task_parse_list_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,20 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('ul');
+const items = document.querySelectorAll('li');
+const arr = [...items];
+
+const sortList = (employees) => {
+  const sorted = (item) => item.dataset['salary'].slice(1).replace(',', '.');
+
+  arr.sort((itemOne, itemTwo) => sorted(itemTwo) - sorted(itemOne));
+
+  return arr;
+};
+
+const getEmployees = (place, employees) => {
+  employees.forEach(element => place.append(element));
+};
+
+sortList(items);
+getEmployees(list, arr);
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,20 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('ul');
+const items = document.querySelectorAll('li');
+const arr = [...items];
+
+const sortList = (employees) => {
+  const sorted = (item) => item.dataset['salary'].slice(1).replace(',', '.');
```
TarasHoliuk, line 8:
```suggestion
  const getSalary = (item) => item.dataset['salary'].slice(1).replace(',', '.');
```
also, you have to compare them as numbers (do type conversion)

TarasHoliuk (changes requested):
Check comment

src/scripts/main.js
```diff
@@ -1,3 +1,22 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('ul');
+const items = document.querySelectorAll('li');
+const arr = [...items];
```
tetiana-ivanchenko, line 5: arr - is not the best name for a variable, it would be better if its name reflected what is inside it.

tetiana-ivanchenko (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,22 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('ul');
+const items = document.querySelectorAll('li');
+const listItems = [...items];
```
AlisaVlasova, line 5:
```suggestion
const listItems = [...document.querySelectorAll('li')];
```

AlisaVlasova (changes requested):

AlisaVlasova (approved):

PR #564, opened 2022-10-06
Solution

SergIats:
[DEMO LINK](https://SergIats.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,25 @@
 'use strict';
 
-// write code here
+const employeesList = [...document.getElementsByTagName('li')]
+  .map(item => [item.innerText, { ...item.dataset }]);
+
+function sortList(list) {
+  list.sort((a, b) => (+b[1].salary.slice(1)
+    .split(',').join('')) - (+a[1].salary.slice(1).split(',').join('')));
+
+  return list;
+};
+
+function getEmployees(list) {
+  const objectArray = [];
+
+  for (const item of list) {
+    objectArray.push(item[1]);
+    objectArray[objectArray.length - 1].name = item[0];
+  };
+
+  return objectArray;
+}
+
+sortList(employeesList);
+getEmployees(employeesList);
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_task_parse_list_DOM/)
+    - [DEMO LINK](https://SergIats.github.io/js_task_parse_list_DOM/)
 2. Follow [this instructions](https://mate-academy.github.io/layout_task-guideline/)
     - Run `npm run test` command to test your code;
     - Run `npm run test:only -- -n` to run fast test ignoring linter;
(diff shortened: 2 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
 'use strict';
 
-// write code here
+const employeeList = [...document.querySelectorAll('li')];
+const mainList = document.querySelector('ul');
+
+function sortList(list) {
+  list.sort((a, b) => (+b.dataset.salary.slice(1)
+    .split(',').join('')) - (+a.dataset.salary.slice(1).split(',').join('')));
```
SanyaBratashchuk, line 8: move normalizing to the separate function

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+
+  for (const item of list) {
+    mainList.append(item);
+  }
+}
```
SanyaBratashchuk, line 13:
```suggestion
    mainList.append(...list);
}
```

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+
+function getEmployees(list) {
+  const objectArray = [];
```
SanyaBratashchuk, line 16:
```suggestion
  const employees = [];
```

SanyaBratashchuk (changes requested):

SergIats:
Hi, Alexandr! Thank you for all recommendations. I have done them.

src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+    employees.push({
+      name: item.innerText,
+      position: item.dataset.position,
+      salary: item.dataset.salary,
+      age: item.dataset.age,
+    });
+  }
+
+  return employees;
+}
```
tetiana-ivanchenko, line 31: why do you have this function? I mean: you create some function that returns new array, and then where do you need that array?

src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
 'use strict';
 
-// write code here
+const employeeList = [...document.querySelectorAll('li')];
+const mainList = document.querySelector('ul');
+
+function salarySort(list) {
+  list.sort((a, b) => (+b.dataset.salary.slice(1).split(',').join(''))
+   - (+a.dataset.salary.slice(1).split(',').join('')));
```
tetiana-ivanchenko, line 8: mb you already have some function that get info from dataset?

tetiana-ivanchenko (changes requested):

src/scripts/main.js
SergIats: Hi, Tetiana! Thanks for review!

I just thought that second function must create exactly new array (I don't know why I thought it)) )

If I have understood you in the right way, the second function must work with the previous list and I don't have to create a new array. I have done it, check please!

src/scripts/main.js
AlisaVlasova: Hi, I see you haven't fixed previous comments, maybe you did not push your changes?
If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review 🙂

AlisaVlasova (changes requested):

tetiana-ivanchenko (approved):

PR #568, opened 2022-10-12
Solution

Fispil:
https://Fispil.github.io/js_task_parse_list_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,28 @@
 'use strict';
 
 // write code here
+const entryList = [...document.querySelectorAll('li')];
+
+function sortList(list) {
+  const result = list
+    .map(item => +(item.dataset.salary.shift().split(',').join('')))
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
 'use strict';
 
 // write code here
+const entryList = [...document.querySelectorAll('li')];
+
+function sortList(list) {
+  const result = list
+    .map(item => +(item.dataset.salary.shift().split(',').join('')))
+    .sort((a, b) => b.datase.salary - a.dataset.salary);
```
Anastasiia-Svintsova, line 9: is it work correctly? Check one more time 
```suggestion
  const result = list
    .map(item => +(item.dataset.salary.shift().split(',').join('')))
    .sort((a, b) => b.dataset.salary - a.dataset.salary);
```

Anastasiia-Svintsova (changes requested):

qoa11a (approved):

PR #578, opened 2022-10-20
solution

MelnychukYana:
https://Melnychuk.github.io/js_task_parse_list_DOM/


DarkMistyRoom (changes requested):
Demo link doesn't work
image
did you run deploy? ask in fe_chat if smth is unclear

PR #593, opened 2022-11-01
solution

MaxFedak:
- [DEMO LINK](https://maxfedak.github.io/js_task_parse_list_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
...
+}
+
+const employeeArray = getEmployees();
+const employeeSorted = sortList(employeeArray, 'salary');
+const listElement = document.querySelector('ul');
+
+listElement.innerHTML = `${employeeSorted.map(item =>
+  `<li data-position=${item.position}
+    data-salary=${item.salary}
+    data-age=${item.age}>${item.name}</li>`)}`.replaceAll(',', '');
```
illiarerun, line 35: please format html properly

illiarerun, line 35: I guess you can add this logic to your sort list method

src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
...
+  }
+
+  return resultArray;
+}
+
+function sortList(array, key) {
+  return array.sort((prev, cur) => cur[key] - prev[key]);
+}
+
+const employeeArray = getEmployees();
```
illiarerun, line 28: > Create function getEmployees, which takes a list property, and returns an array of right objects.
Call both functions.

you should pass list here as a param

src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
...
+
+function getEmployees() {
+  const resultArray = [];
+  const employeeList = document.querySelectorAll('li');
+
+  employeeList.forEach(
+    item => resultArray.push({
+      ...item.dataset, name: item.innerText,
+    })
+  );
```
illiarerun, line 15: better do it with map

src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
...
+
+  for (const employee of resultArray) {
+    employee.salary = convertToNum(employee, 'salary');
+  }
```
illiarerun, line 19: and call this method right inside the map, no need to loop the second time for that

illiarerun (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,43 @@
...
+function getEmployees(list) {
+  const resultArray = [];
+  const employeeList = [...list];
+
+  employeeList.map((item, index) => {
+    resultArray.push({
+      ...item.dataset, name: item.innerText,
+    });
+    resultArray[index].salary = convertToNum(resultArray[index], 'salary');
+  });
```
illiarerun, line 16: do not `push` values to other arrays using map , just create an object on each iteration and SAVE your result of map to variable, like that:
```suggestion
  const employeeList = [...list];

   const resultArray = employeeList.map((item, index) => {
	   return {
	      ...item.dataset,
	      name: item.innerText,
	      salary: convertToNum(item, 'salary')
	    });
  });
```

illiarerun (changes requested):

illiarerun (approved):

PR #595, opened 2022-11-02
Solution

muckbet:


src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
 'use strict';
 
-// write code here
+function toNumber(salary) {
+  return parseFloat(salary.substring(1).replaceAll(',', ''));
+}
+
+function sortList(list) {
+  const lst = document.querySelector(list);
```
TarasHoliuk, line 8:
```suggestion
  const list = document.querySelector(list);
```
don't use short forms

src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
...
+
+  [...lst.children]
+    .sort((a, b) => toNumber(b.dataset.salary) - toNumber(a.dataset.salary))
+    .forEach(node => lst.appendChild(node));
+}
+
+function getEmployees(list) {
```
TarasHoliuk, line 15: you don't use the argument `list`. delete it or use in query selector as in the `sortList` function

TarasHoliuk: tacking into account that you always use `ul` in querySelector you can delete it here and in `sortList` as well

TarasHoliuk (changes requested):
Don't forget about 11, 18 in the [list](https://mate-academy.github.io/layout_task-guideline/)

src/scripts/main.js
muckbet: list is already a function argument, I can't use it

src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
...
+function sortList(list) {
+  const newList = document.querySelector(list);
+
+  [...newList.children]
+    .sort((a, b) => toNumber(b.dataset.salary) - toNumber(a.dataset.salary))
+    .forEach(node => newList.appendChild(node));
+}
+
+function getEmployees(list) {
+  const newList = document.querySelector(list);
```
AlisaVlasova, line 16:
```suggestion
function getEmployees() {
  const list = document.querySelector('ul');
```
я б навіть винесла `const list = document.querySelector('ul'); ` повище в змінну, щоб можна було в обох функціях перевикористати)

AlisaVlasova (changes requested):

AlisaVlasova (approved):

PR #609, opened 2022-11-16
Parse list is solved

sergey-vorobei:
- [DEMO LINK](https://sergey-vorobei.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,29 @@
 'use strict';
 
-// write code here
+const employsList = [...document.querySelectorAll('li')];
+const listParent = document.querySelector('ul');
+
+function sortList(list) {
+  list.sort((a, b) => {
+    return normalize(b.dataset.salary) - normalize(a.dataset.salary);
+  });
+
+  list.forEach(element => listParent.append(element));
+}
+
+function normalize(item) {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+
+sortList(employsList);
+
+function getEmployees(list) {
+  return list.map(item => {
+    return {
+      name: item.innerText,
+      ...item.dataset,
+    };
+  });
```
polosanya, line 26:
```suggestion
  return list.map(item => ({
      name: item.innerText,
      ...item.dataset,
    }));
```

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
 'use strict';
 
-// write code here
+const employsList = [...document.querySelectorAll('li')];
+const listParent = document.querySelector('ul');
+
+function sortList(list) {
+  list.sort((a, b) => {
```
polosanya, line 7: Sort is a mutable method, so it's better to create a copy of list inside the function to be sure that it doesn't mutate array from parameter

polosanya (changes requested):
Almost done! Please provide some fixes

loralevitska (approved):
good job!

PR #614, opened 2022-11-19
Employees list sorting feature was implemented

Andrii-Medintsev:


src/scripts/main.js
```diff
@@ -1,3 +1,26 @@
...
+
+function convertSalary(salaryString) {
+  return salaryString.slice(1).replace(',', '.') * 1000;
+}
+
+function sortList(list) {
+  return [...list.children]
+    .sort((a, b) =>
+      convertSalary(b.dataset.salary) - convertSalary(a.dataset.salary))
+    .forEach(elem => list.append(elem));
```
polosanya, line 13:
```suggestion
    .forEach(element => list.append(element));
```
It is better to use full names, especially if it's about saving only 2-3 chars

src/scripts/main.js
```diff
@@ -1,3 +1,26 @@
...
+}
+
+function getEmployees(list) {
+  return [...list.children].map(employee => ({
```
polosanya, line 17: Map is a non-mutable method, so there is no need to create a copy

polosanya (changes requested):
Almost perfect! Please pay attention to my comments

polosanya (approved):
Well done!

PR #619, opened 2022-11-22
Parse list

maksym-kryvolap:
- [DEMO LINK](https://maksym-kryvolap.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,25 @@
 'use strict';
 
-// write code here
+const names = document.querySelectorAll('li');
+
+function sortList(personList) {
+  return personList.sort((a, b) => b.salary - a.salary);
+}
+
+function getEmployees(people) {
+  return [...people].map(person => {
+    return {
+      name: person.textContent.replace(/\n/g, '').trim(),
+      position: person.dataset.position,
+      salary: +person.dataset.salary.slice(1).split(',').join(''),
+      age: +person.dataset.age,
+    };
+  });
+}
+
+sortList(getEmployees(names));
+
+sortList(getEmployees(names)).forEach((element, index) => {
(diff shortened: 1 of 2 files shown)


tetiana-ivanchenko (changes requested):
fix demo link

src/scripts/main.js
```diff
@@ -1,3 +1,25 @@
...
 
-// write code here
+const names = document.querySelectorAll('li');
+
+function sortList(personList) {
+  return personList.sort((a, b) => b.salary - a.salary);
+}
+
+function getEmployees(people) {
+  return [...people].map(person => {
```
loralevitska, line 10: the `map` method returns a new array, a copy is redundant here
```suggestion
  return people.map(person => {
```

src/scripts/main.js
```diff
@@ -1,3 +1,25 @@
...
+    return {
+      name: person.textContent.replace(/\n/g, '').trim(),
+      position: person.dataset.position,
+      salary: +person.dataset.salary.slice(1).split(',').join(''),
+      age: +person.dataset.age,
+    };
+  });
+}
+
+sortList(getEmployees(names));
```
loralevitska, line 20: fix parameter naming. You pass here `names`, and takes `people` in function. Looks like you you pass array of strings and takes array of object

src/scripts/main.js
```diff
@@ -1,3 +1,25 @@
...
+
+sortList(getEmployees(names)).forEach((element, index) => {
```
loralevitska, line 22: repeated function call from line 20. If you need to process the result of the function, then store it in a variable

loralevitska (changes requested):
almost done!

qoa11a (approved):

PR #621, opened 2022-11-25
solution

divineretribution1:
[DEMO LINK](https://divineretribution1.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,41 @@
 'use strict';
 
-// write code here
+const employeesNodeList = document.querySelectorAll('body > ul > li');
+const ulNode = document.querySelector('body > ul');
+const convertToObjects = (nodeList) => {
+  const arrOfEmployees = [];
+
+  for (const employee of Array.from(nodeList)) {
+    arrOfEmployees.push({
+      name: employee.textContent.trim(),
+      salary: employee.getAttribute('data-salary'),
+      age: employee.getAttribute('data-age'),
+      position: employee.getAttribute('data-position'),
+    });
+  }
+
+  return arrOfEmployees;
+};
+const sortArrBySalary = (arr) => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,41 @@
...
+      age: employee.getAttribute('data-age'),
+      position: employee.getAttribute('data-position'),
+    });
+  }
+
+  return arrOfEmployees;
+};
+const sortArrBySalary = (arr) => {
+  const convertStrToNumber = (str) => Number(str.replaceAll(',', '')
+    .replaceAll('$', ''));
```
TarasHoliuk, line 21:
```suggestion
  const convertStrToNumber = (str) => Number(str.slice(1).replaceAll(',', ''));
```
```suggestion
  const convertStrToNumber = (str) => Number(str.replaceAll(',', '')
    .replaceAll('$', ''));
```
`$` is always in the very beginning

src/scripts/main.js
```diff
@@ -1,3 +1,41 @@
 'use strict';
 
-// write code here
+const employeesNodeList = document.querySelectorAll('body > ul > li');
+const ulNode = document.querySelector('body > ul');
+const convertToObjects = (nodeList) => {
+  const arrOfEmployees = [];
+
+  for (const employee of Array.from(nodeList)) {
```
TarasHoliuk, line 8: It's better to use `map` here: `return Array.from(nodeList).map(// callback)`

TarasHoliuk (changes requested):

src/scripts/main.js
divineretribution1: ok, but what is the purpose of this change? this will make code harder to read

loralevitska (changes requested):
almost done!
please fix eslint error 
image

Anastasiia-Svintsova (approved):

PR #623, opened 2022-11-30
Develop

diolna:
- [DEMO LINK](https://diolna.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,37 @@
 'use strict';
 
-// write code here
+const list = document.getElementsByTagName('li');
+
+function getEmployees(listProperty) {
+  const employees = [];
+
+  for (const employee of list) {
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
 'use strict';
 
-// write code here
+const list = document.getElementsByTagName('li');
+
+function getEmployees(listProperty) {
```
vlad-demchuk, line 5: you take `listProperty` in parameters, but you don't use it?

vlad-demchuk (changes requested):

src/scripts/main.js
diolna: Fixed. Add parametr

diolna:
fixed Add parametr

PR #633, opened 2022-12-06
js_task_parse_list_DOM

SerhiiVovchenko:
- [DEMO LINK](https://SerhiiVovchenko.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -7,24 +7,23 @@ function converting(text) {
   return Number(text.replace(/\D/g, ''));
 }
 
-function sort(list) {
+function sortList(list) {
   const copy = [...list];
   const sortElement = copy.sort((a, b) =>
-    converting(b.dataset.salary) - converting(a.dataset.salary));
+    converting(b.dataset.salary) - converting(a.dataset.salary)
+  );
 
   employeesList.append(...sortElement);
 };
 
 function getEmployeesList(list) {
-  return [...list].map(element => {
-    return {
-      name: element.innerText,
-      position: element.dataset.position,
-      salary: element.dataset.salary,
-      age: element.dataset.age,
-    };
-  });
+  return [...list].map(element => ({
+    name: element.innerText,
+    position: element.dataset.position,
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
...
+const employeesList = document.querySelector('ul');
+
+function converting(text) {
+  return Number(text.replace(/\D/g, ''));
+}
+
+function sort(list) {
+  const copy = [...list];
+  const sortElement = copy.sort((a, b) =>
+    converting(b.dataset.salary) - converting(a.dataset.salary));
```
polosanya, line 13:
```suggestion
  const sortElement = copy.sort((a, b) =>
    converting(b.dataset.salary) - converting(a.dataset.salary)
  );
```

src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
...
+
+function getEmployeesList(list) {
+  return [...list].map(element => {
+    return {
+      name: element.innerText,
+      position: element.dataset.position,
+      salary: element.dataset.salary,
+      age: element.dataset.age,
+    };
+  });
```
polosanya, line 26:
```suggestion
  return [...list].map(element => ({
    name: element.innerText,
    position: element.dataset.position,
    salary: element.dataset.salary,
    age: element.dataset.age,
  }));
```
You can just immediately return an object

src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
...
 
-// write code here
+const employees = document.querySelectorAll('li');
+const employeesList = document.querySelector('ul');
+
+function converting(text) {
+  return Number(text.replace(/\D/g, ''));
+}
+
+function sort(list) {
```
polosanya, line 10: It is not a good idea to call your function the same as array method

polosanya (changes requested):
Almost perfect! Please check the comments

Anastasiia-Svintsova (approved):

PR #644, opened 2022-12-13
solution

Roman-Malysh:


src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
 'use strict';
 
 // write code here
+
+const [...list] = document.querySelectorAll('li');
+
+function sortList(list) {
```
TarasHoliuk, line 7: just rename it to avoid lint error

TarasHoliuk (changes requested):

nazarmatsevych (approved):

PR #645, opened 2022-12-13
Solution

OleksandrOse:
[DEMO LINK](https://OleksandrOse.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,18 @@
 'use strict';
 
 // write code here
+const list = document.querySelector('ul');
+
+const listElements = [...list.children];
+
+const sortedElements = listElements.sort((a, b) => {
+  return (salaryInNumber(b)) - (salaryInNumber(a));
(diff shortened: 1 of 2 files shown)


ViktorHavryliuk (changes requested):
Solution is correct, but please divide your logic into separate functions and call them at the end as it is specified in task:
image
Good job 👍

src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
...
+
+const sortList = arr => {
+  arr.sort((a, b) => {
+    return (convertToNumber(b)) - (convertToNumber(a));
+  });
+
+  arr.forEach(el => list.appendChild(el));
+};
+
+const convertToNumber = value => {
```
TarasHoliuk, line 16: Define the function before you use it next time

TarasHoliuk (approved):
Left 1 comment (just FYI)

PR #653, opened 2022-12-19
solution

webdevw:


polosanya (changes requested):
Please add a Demo link to your PR description

webdevw:
> Please add a Demo link to your PR description

I'm sorry, I'm just a little slowly thinking sometimes

src/scripts/main.js
```diff
@@ -1,3 +1,46 @@
 'use strict';
 
-// write code here
+const liElements = document.querySelectorAll('li');
+const ulElement = document.querySelector('ul');
+const arr = [...liElements];
```
polosanya, line 5: Would be great to clarify what exactly this array is. Use descriptive name

src/scripts/main.js
```diff
@@ -1,3 +1,46 @@
...
+
+function toNumb(salary) {
```
polosanya, line 7:
```suggestion
function convertToNumber(salary) {
```
Let's get used to name function according to these rules: 
image

src/scripts/main.js
```diff
@@ -1,3 +1,46 @@
...
+  return +salary.slice(1).replace(',', '');
+}
+
+const sortList = function(list) {
```
polosanya, line 11: Why do you use two different ways of creating a function here and above? It can be confusing for other developers

src/scripts/main.js
```diff
@@ -1,3 +1,46 @@
...
+[...liElements].forEach(el => el.remove());
+
+ulElement.insertAdjacentHTML('afterbegin', `
+    ${[...newLi].map(person => `
+      ${person.outerHTML}
+    `).join('')}
+`);
+
+function getEmployees(list) {
+  const arrayObj = [...list].map(person => {
```
polosanya, line 28: Map is a non-mutating array, so there is no need to create a copy.
Check yourself

src/scripts/main.js
```diff
@@ -1,3 +1,46 @@
...
+    const clearName = person.innerText
+      .slice(2, person.innerText.lastIndexOf('/n'))
+      .trim();
+
+    const obj = {
```
polosanya, line 33: Let's use more descriptive name here too

polosanya (changes requested):
Almost done!

src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+  return list.sort((a, b) => {
+    return (
+      convertToNumber(b.dataset.salary) - convertToNumber(a.dataset.salary)
+    );
+  });
+};
+
+const newLi = sortList(arrayOfNodeList);
+
+// [...liElements].forEach(el => el.remove());
```
polosanya, line 21: Don't forget to remove comments

polosanya (approved):
Good job!

PR #655, opened 2022-12-22
Solution

Vasyliy22:


loralevitska (changes requested):
Please deploy your solution, and before review check demo_link

qoa11a (approved):

PR #660, opened 2022-12-24
Solution

offsset:


offsset:
https://offsset.github.io/js_task_parse_list_DOM/

src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
 'use strict';
 
-// write code here
+const user = document.querySelectorAll('li');
+
+function sortList(list) {
+  const bod = document.querySelector('body');
```
vlad-demchuk, line 6:
```suggestion
  const body = document.querySelector('body');
```

src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+  const userUl = document.querySelector('ul');
+
+  userUl.remove();
```
vlad-demchuk, line 9: why do you remove it?

src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+
+  const result = [...list].sort((a, b) =>
+    b.dataset.salary.slice(1).split(',').join('')
+     - a.dataset.salary.slice(1).split(',').join(''));
+
+  userUlNew.innerHTML = `
+  <ul>
+  ${result.map(text => `<li> ${text.textContent}</li>`).join('')}
+  </ul>
+  `;
```
vlad-demchuk, line 21: format it

src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+
+  list.forEach((employee) => {
+    const resultUser = {};
+    let nameEmployee = employee.innerText;
+
+    nameEmployee = nameEmployee.replace(/\r|\n/g, '');
+
+    nameEmployee = nameEmployee.replace('        ', '');
+
+    const index = nameEmployee.indexOf(' ');
```
vlad-demchuk, line 36: it's not clear. why do you use replace?

vlad-demchuk (changes requested):
make it simpler
let's follow these instructions under task requirements
image

src/scripts/main.js
offsset: По умолчанию получаю текст с li c пробелами и знаками переноса, пытаюсь их убрать.

src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+  ${text.textContent}</li>`).join('')}
+  `;
+
+  userUl.replaceWith(userUlNew);
+}
+
+function getEmployees(list) {
+  const result = [];
+
+  list.forEach((employee) => {
```
TarasHoliuk, line 28:
```suggestion
  return [...list].map((employee) => {
```
do it in the same way as above

src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+
+    nameEmployee = nameEmployee.replace('        ', '');
+
+    const index = nameEmployee.indexOf(' ');
+
+    resultUser.name = nameEmployee.slice(0, index);
+    resultUser.position = employee.dataset.position;
+    resultUser.salary = employee.dataset.salary;
+    resultUser.age = employee.dataset.age;
+    result.push(resultUser);
```
TarasHoliuk, line 42:
```suggestion
    return {
      name: nameEmployee.slice(0, index),
      // other properties
    }
```
do it in this way (in `map` method instead `forEach`)

src/scripts/main.js
TarasHoliuk: not sure it's necessary here, but ok

TarasHoliuk (changes requested):

TarasHoliuk (approved):
:+1:

PR #673, opened 2023-01-02
solution 1.2 --- 

aliliolek:
https://aliliolek.github.io/js_task_parse_list_DOM/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,81 +9,48 @@
   <body>
     <h1>List of employees</h1>
     <ul>
-      <li
-        data-position="Accountant"
-        data-salary="$162,700"
-        data-age="33"
-      >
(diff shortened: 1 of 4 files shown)


src/index.html
```diff
@@ -9,81 +9,48 @@
   <body>
     <h1>List of employees</h1>
     <ul>
-      <li
-        data-position="Accountant"
-        data-salary="$162,700"
-        data-age="33"
-      >
+      <li data-position="Accountant" data-salary="$162,700"
+       data-age="33">
```
loralevitska, line 13: image
Revert the previous solution. It was true. Fix it everywhere

image

loralevitska (changes requested):
Almost done)

PR #676, opened 2023-01-05
Splotion

kate-matviiv:
https://kate-matviiv.github.io/js_task_parse_list_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -9,8 +9,12 @@ function toNumber(str) {
 }
 
 function sortList(list) {
-  list.sort((a, b) => toNumber(b.dataset.salary)
-  - toNumber(a.dataset.salary));
+  list.sort((item1, item2) => {
+    const { salary: salary1 } = item1.dataset;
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,26 @@
...
+
+const peopleList = [...document.querySelectorAll('li')];
+
+function toNumber(str) {
+  return +str.replace(/[^\d]/g, '');
+}
+
+function sortList(list) {
+  list.sort((a, b) => toNumber(b.dataset.salary)
+  - toNumber(a.dataset.salary));
```
loralevitska, line 13: If such a function does not fit in the line, then it is better to write it in full with desctructure salary from object and rename `a` and `b` to more meaningful variable. 

For what? When everything fits on one line, there is little action and everything is clear, when there is more, then confusion can arise, which we want to avoid, so that our code can be easily read by any developer

loralevitska (changes requested):
Please deploy your solution, and before review check demo_link

polosanya (approved):
Good job!

PR #677, opened 2023-01-05
Solution

yana-kotsulym:
https://yana-kotsulym.github.io/js_task_parse_list_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,27 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('ul');
+
+const sortList = (listOfValue) => {
+  let sort = [...listOfValue].sort((a, b) =>
+    b.dataset.salary.split(',').join('').slice(1)
+    - a.dataset.salary.split(',').join('').slice(1));
+
+  sort = sort.forEach(item => list.append(item));
+
+  return sort;
+};
+
+function getEmployees(listOfValue) {
+  const result = [...listOfValue].map(item => ({
+    name: item.textContent.trim(),
+    position: item.dataset.position,
+    salary: item.dataset.salary,
+    age: item.dataset.age,
+  }));
+
+  return result;
+}
+
+sortList(list.children);
+getEmployees(list.children);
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,27 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('ul');
+
+const sortList = (listOfValue) => {
+  let sort = [...listOfValue].sort((a, b) =>
+    b.dataset.salary.split(',').join('').slice(1)
+    - a.dataset.salary.split(',').join('').slice(1));
```
SanyaBratashchuk, line 8: create separate function for normalizing number

SanyaBratashchuk (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
...
+const sortList = (listOfValue) => {
+  const listWithNumber = [...listOfValue];
+
+  function normalizeNumber() {
+    for (const item of listWithNumber) {
+      item.dataset.salary = item.dataset.salary.split(',').join('').slice(1);
+    }
+  };
+
+  normalizeNumber();
```
ViktorHavryliuk, line 14: It's better to make function just to convert string, and then use it in loop on each value. This vay it's more flexible and better corresponds to its name
```suggestion
  function normalizeNumber(numberString) {
    return numberString.split(',').join('').slice(1);
  };

  for (const item of listWithNumber) {
      item.dataset.salary = normalizeNumber(item.dataset.salary);
  }
```

src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
...
+
+  let sort = listWithNumber.sort((a, b) => b.dataset.salary - a.dataset.salary);
+
+  sort = sort.forEach(item => list.append(item));
+
+  return sort;
+};
+
+function getEmployees(listOfValue) {
+  const result = [...listOfValue].map(item => ({
```
ViktorHavryliuk, line 24: map doesn't mutate original array and returns a new one anyway, so you don't need to copy listOfValues here

ViktorHavryliuk (changes requested):

PR #682, opened 2023-01-07
solution

Avichez:
[DEMO LINK](https://Avichez.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -12,78 +12,67 @@
       <li
         data-position="Accountant"
         data-salary="$162,700"
-        data-age="33"
-      >
+        data-age="33">
         Airi Satou
       </li>
       <li
         data-position="Javascript Developer"
         data-salary="$183,000"
-        data-age="29"
-      >
+        data-age="29">
         Michael Bruce
       </li>
       <li
         data-position="Marketing Designer"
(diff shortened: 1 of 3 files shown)


NVyrsta (approved):
Good job)

src/index.html
```diff
@@ -12,78 +12,67 @@ <h1>List of employees</h1>
       <li
         data-position="Accountant"
         data-salary="$162,700"
-        data-age="33"
-      >
+        data-age="33">
```
Denchepornyuk, line 15: it isn't a correct fix for tag attributes

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+
+    return salaryB - salaryA;
+  });
+
+  listEmployees.append(...sortedList);
+};
+
+const getEmployees = (list) => {
+  return [...list.children].map(employee => {
+    return {
```
Denchepornyuk, line 22: you can avoid return in these arrow functions

Denchepornyuk (approved):

src/index.html
Avichez: Sorry, that was autoformatting probably, observed it only after push.

PR #692, opened 2023-01-14
Solution

MrFox32:


src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
...
+const employeesList = document.querySelector('ul');
+
+const sortListItems = (list) => {
+  const listChildren = [...list.children];
+
+  return listChildren.sort((a, b) => {
+    const innerA = +a.attributes.getNamedItem('data-salary')
+      .value.slice(1).split(',').join('');
+    const innerB = +b.attributes.getNamedItem('data-salary')
+      .value.slice(1).split(',').join('');
```
Anastasiia-Svintsova, line 12: You do the same in both cases. Would be great to create a method for it

src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
...
+
+    if (innerA < innerB) {
+      return 1;
+    } else {
+      return -1;
+    }
```
Anastasiia-Svintsova, line 18: Just return the result of subtracting b from a

Anastasiia-Svintsova (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,24 @@
 'use strict';
 
-// write code here
+const employeesList = document.querySelector('ul');
+
+const sortListItems = (list) => {
+  const listChildren = [...list.children];
+
+  return listChildren.sort((a, b) => getSalaryValue(b) - getSalaryValue(a));
```
Anastasiia-Svintsova, line 8: It's better to use functions and variables after their declaration
```
const getSalaryValue = ...

const sortListItems = ...

Anastasiia-Svintsova (approved):
👍

PR #694, opened 2023-01-15
Solution

andrieieva-yuliia:
- [DEMO LINK](https://andrieieva-yuliia.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,28 @@
 'use strict';
 
-// write code here
+const peopleOfList = document.querySelectorAll('li');
+
+const listString = function(string) {
+  return +string.slice(1).split(',').join('');
+};
+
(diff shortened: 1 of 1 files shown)


NVyrsta (approved):
Good job

src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
 'use strict';
 
-// write code here
+const peopleOfList = document.querySelectorAll('li');
+
+const listString = function(string) {
+  return +string.slice(1).split(',').join('');
+};
```
Anna-Tkachenko, line 7: will be better to use such syntax for object methods
if you want to create function and save it to the variable -> use arrow function
for other cases you can use function declaration and specify name after the `function` keyword

Anna-Tkachenko (approved):
Good job!

PR #703, opened 2023-01-20
Finish task

U-Ra-All:


src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
...
+function getEmployees(list) {
+  return [...list].map((employee) => {
+    return {
+      name: employee.textContent.trim(),
+      position: employee.dataset.position,
+      salary: employee.dataset.salary,
+      age: employee.dataset.age,
+    };
+  });
+}
```
IvanFesenko, line 34: Useless destructuring, map method return new Array
```suggestion
function getEmployees(list) {
  return list.map(({textContent, dataset}) => ({
    name: textContent.trim(),
    position: dataset.position,
    salary: dataset.salary,
    age: dataset.age,
    }));
}
```

src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
...
+
+  employees.sort(
+    (a, b) => convertSalaryToNumber(b.salary) - convertSalaryToNumber(a.salary)
+  );
+
+  for (let i = 0; i < employees.length; i++) {
+    list[i].dataset.position = employees[i].position;
+    list[i].dataset.salary = employees[i].salary;
+    list[i].dataset.age = employees[i].age;
+    list[i].textContent = employees[i].name;
```
IvanFesenko, line 17: Do not mutate current dom tree, apent new children

IvanFesenko (changes requested):
Some fix require

src/scripts/main.js
U-Ra-All: I get an array from NodeList by destructuring

IvanFesenko: Agreed, miss that

src/scripts/main.js
U-Ra-All: Ok, done

PR #709, opened 2023-01-25
solution

vladyslav-krainyk:
[DEMO LINK](https://vladyslav-krainyk.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,29 @@
 'use strict';
 
-// write code here
+const listOfEmployees = document.querySelector('ul');
+const arrOfEmployees = [...document.querySelectorAll('li')];
+
+function convertToNumber(value) {
+  return +value.slice(1).split(',').join('');
+}
+
+function sortList(item) {
+  const sortItem = item.sort((a, b) =>
+    convertToNumber(b.dataset.salary) - convertToNumber(a.dataset.salary)
+  );
+
+  listOfEmployees.append(...sortItem);
+};
+
+sortList(arrOfEmployees);
+
+function getEmployees(item) {
+  return item.map(el => ({
+      name: el.textContent,
+      position: el.dataset.position,
+      salary: el.dataset.salary,
+      age: el.dataset.age,
+  }));
+}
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+  return +value.slice(1).split(',').join('');
+}
+
+function sortList(item) {
+  const sortItem = item.sort((a, b) =>
+    convertToNumber(b.dataset.salary) - convertToNumber(a.dataset.salary)
+  );
+
+  listOfEmployees.append(...sortItem);
+};
```
DarkMistyRoom, line 16: `sortList `function should ONLY sort the list! It is important. Everyone should understand what the function does from its name

DarkMistyRoom (changes requested):

ViktorHavryliuk (changes requested):
Feedback of previous mentor is not implemented. Check if you have pushed your changes, and feel free to ask in chat if you have questions

src/scripts/main.js
```diff
@@ -1,3 +1,24 @@
...
+const arrOfEmployees = [...document.querySelectorAll('li')];
+
+const convertToNumber = value => +value.slice(1).split(',').join('');
+
+function sortList(list) {
+  return list.sort((a, b) => convertToNumber(b.dataset.salary)
+    - convertToNumber(a.dataset.salary));
+}
+
+listOfEmployees.append(...sortList(arrOfEmployees));
```
Volodymyr-Melnyk-Fron10der, line 13: It works, but there is no reason to call `sortList `like this, because sort is a mutable method and changes the initial array
```suggestion
function sortList(list) {
  return list.sort((a, b) => convertToNumber(b.dataset.salary)
    - convertToNumber(a.dataset.salary));
}

sortList(arrOfEmployees)

listOfEmployees.append(...arrOfEmployees);
```

Volodymyr-Melnyk-Fron10der (approved):
Good job! Buk keep in mind that sort is a mutable method

PR #715, opened 2023-02-01
Parse list DOM

ruslan-yarosh:
[DEMO LINK](https://ruslan-yarosh.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,28 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('ul');
+const listItem = document.querySelectorAll('li');
+
+function sortList(listToSort) {
+  const sortedList = [...listToSort].sort((a, b) => {
+    const aFormatted = formatSalary(a.getAttribute('data-salary'));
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
...
+
+function sortList(listToSort) {
+  const sortedList = [...listToSort].sort((a, b) => {
+    const aFormatted = formatSalary(a.getAttribute('data-salary'));
+    const bFormatted = formatSalary(b.getAttribute('data-salary'));
+
+    return bFormatted - aFormatted;
+  });
+
+  return sortedList.map(el => list.append(el));
```
ViktorHavryliuk, line 14: Better to use forEach in this case, as you only iterating through those elements to append them, you don't need return here too.
```suggestion
 sortedList.forEach(el => list.append(el));
```
Map is used when you need to create an array of new objects based on exesting ones to use that new array somehow

ViktorHavryliuk (changes requested):

tetiana-ivanchenko (approved):

PR #723, opened 2023-02-08
Task done

petrolozynskyi:


src/scripts/main.js
```diff
@@ -1,3 +1,38 @@
...
 
-// write code here
+const nodes = document.querySelectorAll('li');
+const list = document.querySelector('ul');
+
+const frag = document.createDocumentFragment();
+
+const sortedList = Array.from(nodes).sort((a, b) =>
+  parseInt(a.dataset.salary.replace(/\D/g, ''))
+- parseInt(b.dataset.salary.replace(/\D/g, '')));
```
Volodymyr-Melnyk-Fron10der, line 10: It`s better to move this variable into sortList function, because it belongs to sortList

Volodymyr-Melnyk-Fron10der (changes requested):
Great job! Just one comment  here

src/scripts/main.js
petrolozynskyi: Thanks, but have a next problem if I move this into sortListFunction:
The function "getEmployees()" use this const too, so for fixing your comment, I just move "sortedList" into two of these functions

src/scripts/main.js
```diff
@@ -1,3 +1,41 @@
 'use strict';
 
-// write code here
+const nodes = document.querySelectorAll('li');
+const list = document.querySelector('ul');
+
+const frag = document.createDocumentFragment();
+
+const sortList = function(someList) {
```
slavanosatskyi, line 8:
```suggestion
function sortList(someList) {
```

slavanosatskyi, line 8: why do you need someList argument, seems that you do not use it

slavanosatskyi, line 8: Your function is doing too much, it is not only sorting data but also manipulates the DOM. you can split it into two functions. More over, one of them you will be able to use in getEmployees function.

slavanosatskyi (changes requested):
Feel free to contact mentors in corresponding chats if something is not clear

ViktorHavryliuk (approved):

PR #725, opened 2023-02-09
Solution

Vasyl-Pavlenko:
[DEMO LINK](https://vasyl-pavlenko.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,28 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('ul');
+const employees = [...document.querySelectorAll('li')];
+
+function sortEmployees(arr) {
+  const salary = (item) => +item.slice(1).split(',').join('');
+
+  const sorted = arr.sort((a, b) => {
+    return salary(b.dataset.salary) - salary(a.dataset.salary);
+  });
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
...
+
+function getEmployees(arr) {
+  return arr.map((item) => {
+    return {
+      name: item.innerText,
+      position: item.dataset.position,
+      salary: item.dataset.salary,
+      age: item.dataset.age,
+    };
+  });
```
IvanFesenko, line 24: Destructuring make it more clear
```suggestion
  return arr.map(({innerText, dataset: {position, salary, age}}) => {
    return {
      name,
      position,
      salary,
      age,
    };
  });
```

src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('ul');
+const employees = [...document.querySelectorAll('li')];
+
+function sortEmployees(arr) {
+  const salary = (item) => +item.slice(1).split(',').join('');
+
```
IvanFesenko, line 8: It should be outside current function

PR #729, opened 2023-02-10
Solution

IrynaK-a:
https://IrynaK-a.github.io/js_task_parse_list_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,27 @@
 'use strict';
 
-// write code here
+const listOfEmployees = document.querySelector('ul');
+let items = document.querySelectorAll('li');
+
+function sortList(list) {
+  listOfEmployees.innerHTML = '';
+
+  const sortedList = [...list].sort((a, b) => {
+    const salaryA = Number(a.dataset.salary.slice(1).split(',').join(''));
+    const salaryB = Number(b.dataset.salary.slice(1).split(',').join(''));
+
+    return salaryB - salaryA;
+  });
+
+  sortedList.forEach((item) => listOfEmployees.append(item));
+  items = document.querySelectorAll('li');
+}
+
+function getEmployees(list) {
+  const sortedArray = [...list];
+
+  return sortedArray;
+}
+
+sortList(items);
+getEmployees(items);


src/scripts/main.js
```diff
@@ -1,3 +1,27 @@
...
-// write code here
+const listOfEmployees = document.querySelector('ul');
+let items = document.querySelectorAll('li');
+
+function sortList(list) {
+  listOfEmployees.innerHTML = '';
+
+  const sortedList = [...list].sort((a, b) => {
+    const salaryA = Number(a.dataset.salary.slice(1).split(',').join(''));
+    const salaryB = Number(b.dataset.salary.slice(1).split(',').join(''));
```
Anastasiia-Svintsova, line 11: Look, you do the same actions several times. Would be great to create a method for converting to a number

src/scripts/main.js
```diff
@@ -1,3 +1,27 @@
 'use strict';
 
-// write code here
+const listOfEmployees = document.querySelector('ul');
+let items = document.querySelectorAll('li');
+
+function sortList(list) {
+  listOfEmployees.innerHTML = '';
```
Anastasiia-Svintsova, line 7: Why do you need this? Seems that it's redundant  here 🤔

src/scripts/main.js
```diff
@@ -1,3 +1,27 @@
...
+}
+
+function getEmployees(list) {
+  const sortedArray = [...list];
+
+  return sortedArray;
+}
+
+sortList(items);
+getEmployees(items);
```
Anastasiia-Svintsova, line 27: Same about the getEmployees function. Why do you need it?
```suggestion
sortList(items);
```

Anastasiia-Svintsova (changes requested):

src/scripts/main.js
IrynaK-a: > Same about the getEmployees function. Why do you need it?

I did it because it is in requirement "
3) Create function `getEmployees`, which takes a `list` property, and returns an array of right objects.
4) Call both functions.
"

witflash (approved):
Well done 👍

PR #731, opened 2023-02-13
done

l4st1m0za:


src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
 'use strict';
 
-// write code here
+const employeesList = document.querySelector('ul');
+const employees = [...document.querySelectorAll('li')];
+
+function salaryNum(salary) {
```
slavanosatskyi, line 6: functions should be named with verbs - convertSalaryToNumber, parseSalaryValue, etc...

slavanosatskyi (changes requested):
you did a good work, only two minor comments to be fixed. And don't forget to add link to your demo in pr's description :)

PR #735, opened 2023-02-14
Chupiak

Chupiak:


src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
 'use strict';
 
-// write code here
+// eslint-disable-next-line no-shadow
```
tetiana-ivanchenko, line 3: remove all eslint-disable comments

tetiana-ivanchenko (changes requested):

src/scripts/main.js
Chupiak: @tetiana-ivanchenko Done. Please check.

src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
...
+      age: employee.getAttribute('data-age'),
+    };
+  });
+}
+
+function converter(salary) {
+  return salary.replaceAll(',', '').replace('$', '');
+}
+
+const worker = [...document.body.children[1].children];
```
ViktorHavryliuk, line 27: Would be better to access workers via getting ul element and its children, rather then just children chain. This way it's more clear what you are looking for
However it's not crucial

ViktorHavryliuk (changes requested):
Please fix demo link not being shown

src/scripts/main.js
Chupiak: @ViktorHavryliuk Done.

ViktorHavryliuk (approved):

PR #742, opened 2023-02-20
Solution

taras-nakonechnyi:
https://taras-nakonechnyi.github.io/js_task_parse_list_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,23 @@
 'use strict';
 
-// write code here
+const workerColection = Array.from(
+  document.getElementsByTagName('li')
+);
+const listWorker = document.querySelector('ul');
+const solaryArray = [];
+
+workerColection.forEach(element => {
+  solaryArray.push(formatItem(element));
+});
+
+function formatItem(item) {
+  return Number(item.dataset.salary.match(/\d+/g).join(''));
+}
+
+solaryArray.sort((a, b) => b - a).map(salary => {
+  workerColection.forEach(element => {
+    if (formatItem(element) === salary) {
+      listWorker.append(element);
+    };
+  });
+});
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,23 @@
...
+
+workerColection.forEach(element => {
+  solaryArray.push(formatItem(element));
+});
+
+function formatItem(item) {
+  return Number(item.dataset.salary.match(/\d+/g).join(''));
+}
+
+solaryArray.sort((a, b) => b - a).map(salary => {
```
tetiana-ivanchenko, line 17: if you shouldn't return new array - use forEach instead of map

src/scripts/main.js
```diff
@@ -1,3 +1,23 @@
 'use strict';
 
-// write code here
+const workerColection = Array.from(
+  document.getElementsByTagName('li')
+);
+const listWorker = document.querySelector('ul');
+const solaryArray = [];
```
tetiana-ivanchenko, line 7:
```suggestion
const saqlaries = [];
```

src/scripts/main.js
```diff
@@ -1,3 +1,23 @@
 'use strict';
 
-// write code here
+const workerColection = Array.from(
```
tetiana-ivanchenko, line 3:
```suggestion
const workers = Array.from(
```

tetiana-ivanchenko (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,23 @@
 'use strict';
 
-// write code here
+const workers = Array.from(
+  document.getElementsByTagName('li')
+);
+const listWorker = document.querySelector('ul');
+const solaries = [];
```
nazarmatsevych, line 7:
```suggestion
const salaries= [];
```

nazarmatsevych (approved):
Good work,
just a small note

PR #743, opened 2023-02-21
Solution

dissolveyo:
- [DEMO LINK](https://dissolveyo.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,45 @@
 'use strict';
 
-// write code here
+const ulList = [...document.querySelector('ul').children];
+const newList = document.querySelector('ul');
+
+function formatSalary(employee) {
+  const salary = employee.dataset.salary;
+  const formattedSalary = +salary.replaceAll(',', '').slice(1);
+
+  return formattedSalary;
+}
+
+function sortList(list) {
+  list.sort(
+    (firstEmployee, secondEmployee) =>
+      formatSalary(secondEmployee) - formatSalary(firstEmployee)
+  );
+
+  newList.innerHTML = `
+  ${list.map((element) => `<li>${element.textContent}</li>`).join('')}`;
+}
+
+function getEmployees(list) {
+  const arrayOfEmployees = [];
+
+  const employeeData = list.map((element) => element.dataset);
+
+  for (let i = 0; i < employeeData.length; i++) {
+    const employeeObject = {};
+    const trimmedEmployeeName = list[i].innerText.trim();
+
+    employeeObject.name = trimmedEmployeeName;
+
+    for (const key in employeeData[i]) {
+      employeeObject[key] = employeeData[i][key];
+    }
+
+    arrayOfEmployees.push(employeeObject);
+  }
+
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
...
+
+function formatSalary(employee) {
+  const salary = employee.dataset.salary;
+  const formattedSalary = +salary.replaceAll(',', '').slice(1);
+
+  return formattedSalary;
+}
+
+function sortList(list) {
+  list.sort(
```
Volodymyr-Melnyk-Fron10der, line 14: Mutate arguments it's a bad practice, try to avoid it, make a copy

Volodymyr-Melnyk-Fron10der (changes requested):
I've just one remark here

dissolveyo:
I added return value for sortList function, because If we didn't do that i assume, that we will lose our new sortedList in memory.

src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
 'use strict';
 
-// write code here
+const ulList = [...document.querySelector('ul').children];
```
Viktor-Kost, line 3: it's a good practice to add ref/el suffix to variables with a reference to a node element
```suggestion
const ulListRef = [...document.querySelector('ul').children];
```

src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+const newList = document.querySelector('ul');
+
+function formatSalary(employee) {
```
Viktor-Kost, line 6: it's better if your function accept salary as an argument. It will be more clear

src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+  const salary = employee.dataset.salary;
+  const formattedSalary = +salary.replaceAll(',', '').slice(1);
+
+  return formattedSalary;
```
Viktor-Kost, line 10: then if you pas only salary you can just write only one line of code inside function body
```suggestion
return  +salary.replaceAll(',', '').slice(1);
```

slavanosatskyi (changes requested):
seems that something is broken, as expected result is not met
Expected: 
Screenshot 2023-02-22 at 17 21 37
Actual:
Screenshot 2023-02-22 at 17 22 01

src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+    employeeObject.name = trimmedEmployeeName;
+
+    for (const key in employeeData[i]) {
+      employeeObject[key] = employeeData[i][key];
+    }
+
+    arrayOfEmployees.push(employeeObject);
+  }
+
+  return arrayOfEmployees;
```
Viktor-Kost, line 45: you can simplify your function. 
Just use a `map` method to your `list` and create an object inside 
```
  {
    fullname: elem.innerText,
    position: elem.dataset.position,
    salary: elem.dataset.salary,
    age: +elem.dataset.age,
  }
```

Viktor-Kost (changes requested):

PR #763, opened 2023-03-02
solution

toljan1:
[DEMO LINK](https://toljan1.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,24 @@
 'use strict';
 
-// write code here
+let list = document.querySelectorAll('li');
+
+function sortList(someList) {
+  list = [...someList].sort((person1, person2) =>
+    +person2.dataset.salary.slice(1).split(',').join('')
+      - +person1.dataset.salary.slice(1).split(',').join(''));
+
+  document.querySelector('ul').append(...list);
+}
+
+function getEmployees(someList) {
+  return [...someList].map(person => {
+    return {
+      name: person.innerText,
+      ...person.dataset,
+    };
+  });
+}
+
+sortList(list);
+
+getEmployees(list);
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,24 @@
...
+  document.querySelector('ul').append(...list);
+}
+
+function getEmployees(someList) {
+  return [...someList].map(person => {
+    return {
+      name: person.innerText,
+      ...person.dataset,
+    };
+  });
```
etojeDenys, line 19: can omit
```suggestion
  return [...someList].map(person => ({
      name: person.innerText,
      ...person.dataset,
    })
  );
```

src/scripts/main.js
```diff
@@ -1,3 +1,24 @@
 'use strict';
 
-// write code here
+let list = document.querySelectorAll('li');
+
+function sortList(someList) {
+  list = [...someList].sort((person1, person2) =>
+    +person2.dataset.salary.slice(1).split(',').join('')
+      - +person1.dataset.salary.slice(1).split(',').join(''));
```
etojeDenys, line 8: let`s move it to some variables to improve sort function readability

src/scripts/main.js
```diff
@@ -1,3 +1,26 @@
...
+function getEmployees(someList) {
+  return [...someList].map(person => ({
+    name: person.innerText,
+    ...person.dataset,
+  }));
+}
+
+sortList(list);
+
+getEmployees(list);
```
OlEzhA0, line 26: I really can't understand why do we need this function? you just getting new array with `name` property but it's never used

src/scripts/main.js
```diff
@@ -1,3 +1,26 @@
 'use strict';
 
-// write code here
+let list = document.querySelectorAll('li');
+
+function toNumber(str) {
+  return +str.dataset.salary.slice(1).split(',').join('');
+}
```
OlEzhA0, line 7:
```suggestion
function toNumber(str) {
  return +str.dataset.salary.slice(1).replaceAll(',', '');
}
```

OlEzhA0 (changes requested):

IvanFesenko (approved):

PR #765, opened 2023-03-03
Parse list

Yevhen-Ivanchuk:
- [DEMO LINK](https://Yevhen-Ivanchuk.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,24 @@
 'use strict';
 
-// write code here
+const list = document.querySelectorAll('li');
+const number = (str) => {
+  return Number(str.slice(1).split(',').join(''));
+};
+
+const salary = (employees) => {
+  return [...employees].sort((a, b) =>
+    number(b.dataset.salary) - number(a.dataset.salary));
+};
+
+const getEmployees = (employees) => {
+  return [...employees].map(person => (
+    {
+      name: person.innerText,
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,24 @@
 'use strict';
 
-// write code here
+const list = document.querySelectorAll('li');
+const number = (str) => {
```
witflash, line 4: It's better to name smth like `getNumber` or `parseNumber` for example

src/scripts/main.js
```diff
@@ -1,3 +1,24 @@
...
+  return Number(str.slice(1).split(',').join(''));
+};
+
+const salary = (employees) => {
```
witflash, line 8: And this: `getSalary`

src/scripts/main.js
```diff
@@ -1,3 +1,24 @@
...
+  return [...employees].sort((a, b) =>
+    number(b.dataset.salary) - number(a.dataset.salary));
+};
+
+const getEmployees = (employees) => {
+  return [...employees].map(person => (
```
witflash, line 14: There is no need to use the spread operator here because the `map` method doesn't mutate an array but returns a new array.

witflash (approved):
Good job!
I left some uncritical comments for you.

PR #784, opened 2023-03-15
Solution

iryna-serniuk:
- [DEMO LINK](https://iryna-serniuk.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,31 @@
 'use strict';
 
-// write code here
+const ul = document.querySelector('ul');
+const li = [...document.querySelectorAll('li')];
+
+function sortList(list) {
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,31 @@
 'use strict';
 
-// write code here
+const ul = document.querySelector('ul');
+const li = [...document.querySelectorAll('li')];
+
+function sortList(list) {
+  const sorted = list.sort((a, b) =>
```
OlEzhA0, line 7: would be better to create array copy in the sort List function
like
```
function sortList(list) {
  const copy = [...list]
  const sorted = copy.sort((a, b) =>
```

OlEzhA0 (approved):

PR #785, opened 2023-03-15
Solution

oleksandr-plias:
[DEMO LINK](https://oleksandr-plias.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,27 @@
 'use strict';
 
-// write code here
+const li = document.querySelectorAll('li');
+
+const sortList = (list) => {
+  function getNumber(salary) {
+    return +salary.slice(1).split(',').join('');
+  }
+
+  const sortedList = [...list].sort((a, b) => {
+    const aSalary = getNumber(a.dataset.salary);
+    const bSalary = getNumber(b.dataset.salary);
+
+    return bSalary - aSalary;
+  });
+
+  const ul = document.querySelector('ul');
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,27 @@
...
+  });
+
+  const ul = document.querySelector('ul');
+
+  sortedList.map(item => ul.appendChild(item));
+};
+
+const getEmployees = (list) => {
+  return [...list].map(item => item.dataset.name);
+};
```
Volodymyr-Melnyk-Fron10der, line 24: not just array of names, but objects
image

src/scripts/main.js
```diff
@@ -1,3 +1,27 @@
 'use strict';
 
-// write code here
+const li = document.querySelectorAll('li');
```
Volodymyr-Melnyk-Fron10der, line 3:
```suggestion
const listItems = document.querySelectorAll('li');
```

src/scripts/main.js
```diff
@@ -1,3 +1,27 @@
...
+
+const sortList = (list) => {
+  function getNumber(salary) {
+    return +salary.slice(1).split(',').join('');
+  }
```
Volodymyr-Melnyk-Fron10der, line 8: move helper functions on global level

Volodymyr-Melnyk-Fron10der (changes requested):
Almost there!

PR #787, opened 2023-03-16
Develop

StetsenkoViktoriia:
[DEMO LINK](https://StetsenkoViktoriia.github.io/js_task_parse_list_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,31 @@
...
+  });
+
+  document.querySelector('ul').append(...list);
+};
+
+sortList(listItems);
+
+function getEmployees(list) {
+  list.map(item => {
+    return {
```
witflash, line 22: And here you can use implicit return too.

It's not a mistake. It's just code optimization

PR #796, opened 2023-03-21
Solution

nazarkharkevych:
https://nazarkharkevych.github.io/js_task_parse_list_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,35 @@
 'use strict';
 
-// write code here
+const listNode = document.querySelector('ul');
+const [...employees] = listNode.children;
+
+sortList(employees);
+getEmployees(employees);
+
+function sortList(list) {
+  list.sort((a, b) => {
+    const aSalary = convertToNumber(a.getAttribute('data-salary'));
+    const bSalary = convertToNumber(b.getAttribute('data-salary'));
+
+    return bSalary - aSalary;
+  });
+
+  listNode.append(...employees);
+};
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
-// write code here
+const listNode = document.querySelector('ul');
+const [...employees] = listNode.children;
+
+sortList(employees);
+listNode.append(...employees);
+getEmployees(employees);
+
+function sortList(list) {
+  list.sort((a, b) => {
```
OlEzhA0, line 11: would be better yo copy `list` here
like
`return [...list].sort...`

and use like

`listNode.append(sortList(employees))`

src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+
+function getEmployees(list) {
+  return list.map((person) => {
+    return {
+      name: person.innerText,
+      position: person.getAttribute('data-position'),
+      salary: person.getAttribute('data-salary'),
+      age: person.getAttribute('data-age'),
+    };
+  });
```
OlEzhA0, line 33:
```suggestion
  return list.map((person) => ({
    name: person.innerText,
    position: person.getAttribute('data-position'),
    salary: person.getAttribute('data-salary'),
    age: person.getAttribute('data-age'),
  }));
```

OlEzhA0 (changes requested):
hey! just added minor improvements

PR #797, opened 2023-03-22
List was parsed and sorted

serge-kontakt:
- [DEMO LINK](https://serge-kontakt.github.io/js_task_parse_list_DOM/)
   
I did not understand what means "'list' property" in the task, so I did the task the way I'd decided.

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,55 @@
 'use strict';
 
-// write code here
+const arrayLike = document.querySelectorAll('li');
+
+const arrayReal = [];
+
+function getEmployees() {
+  for (let i = 0; i < arrayLike.length; i++) {
+    const item = arrayLike[i].dataset;
+    let itemReal = { ...item };
+
+    itemReal = {
+      name: arrayLike[i].innerText, ...itemReal,
+    };
+
+    arrayReal.push(itemReal);
+  }
+
+  return arrayReal;
+}
+
+getEmployees();
+
+const arrayNew = [];
+
+function sortList() {
+  for (let i = 0; i < arrayReal.length; i++) {
+    arrayNew.push({ ...arrayReal[i] });
+  }
+
+  for (let i = 0; i < arrayNew.length; i++) {
+    arrayNew[i].salary = +arrayNew[i].salary.replace(/\D/g, '');
+  }
+
+  arrayNew.sort((a, b) => b.salary - a.salary);
+
+  for (let i = 0; i < arrayNew.length; i++) {
+    arrayNew[i].salary = '$' + arrayNew[i].salary.toLocaleString('en-US');
+
+    for (let j = 0; j < arrayReal.length; j++) {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,55 @@
...
+        document.querySelectorAll('li')[i].dataset.position
+        = arrayNew[i].position;
+
+        document.querySelectorAll('li')[i].dataset.salary = arrayNew[i].salary;
+
+        document.querySelectorAll('li')[i].dataset.age = arrayNew[i].age;
+      }
+    }
+  }
+}
```
DarkMistyRoom, line 53: both of these functions should take `list `as a property
```suggestion
function getEmployees(list) {
  for (let i = 0; i < arrayLike.length; i++) {
    const item = arrayLike[i].dataset;
    let itemReal = { ...item };

    itemReal = {
      name: arrayLike[i].innerText, ...itemReal,
    };

    arrayReal.push(itemReal);
  }

  return arrayReal;
}

getEmployees();

const arrayNew = [];

function sortList(list) {
  for (let i = 0; i < arrayReal.length; i++) {
    arrayNew.push({ ...arrayReal[i] });
  }

  for (let i = 0; i < arrayNew.length; i++) {
    arrayNew[i].salary = +arrayNew[i].salary.replace(/\D/g, '');
  }

  arrayNew.sort((a, b) => b.salary - a.salary);

  for (let i = 0; i < arrayNew.length; i++) {
    arrayNew[i].salary = '$' + arrayNew[i].salary.toLocaleString('en-US');

    for (let j = 0; j < arrayReal.length; j++) {
      if (arrayReal[j].name === arrayNew[i].name) {
        document.querySelectorAll('li')[i].innerText = arrayNew[i].name;

        document.querySelectorAll('li')[i].dataset.position
        = arrayNew[i].position;

        document.querySelectorAll('li')[i].dataset.salary = arrayNew[i].salary;

        document.querySelectorAll('li')[i].dataset.age = arrayNew[i].age;
      }
    }
  }
}
```

src/scripts/main.js
```diff
@@ -1,3 +1,55 @@
...
+  for (let i = 0; i < arrayLike.length; i++) {
+    const item = arrayLike[i].dataset;
+    let itemReal = { ...item };
+
+    itemReal = {
+      name: arrayLike[i].innerText, ...itemReal,
+    };
+
+    arrayReal.push(itemReal);
+  }
```
DarkMistyRoom, line 17: you can just use method `map` to iterate by list and return the array of objects

src/scripts/main.js
DarkMistyRoom, line 53: no need to make so complicated logic. you can just 
`[...list].sort(...)` and then do this
image

DarkMistyRoom (changes requested):
Good, but let's simplify

src/scripts/main.js
serge-kontakt: Hello, what "list" is? What is inside the "list"?

PR #799, opened 2023-03-22
Solved

silcki:


src/scripts/main.js
```diff
@@ -1,3 +1,43 @@
...
+    return {
+      'name': x.textContent.replace(/[\n\r]+|[\s]{2,}/g, ' ').trim(),
+      'position': x.dataset.position,
+      'salary': Number(x.dataset.salary.replace(/[^0-9.-]+/g, '')),
+      'age': +x.dataset.age,
+    };
+  });
+}
+
+sortList(document.getElementsByTagName('ul')[0]);
```
Viktor-Kost, line 43: you can just use a variable both for getEmployees and sortList

```suggestion
const employees = [...document.querySelectorAll('li')];

function getEmployees(list) {
  return list.map((x) => {
    return {
      'name': x.textContent.replace(/[\n\r]+|[\s]{2,}/g, ' ').trim(),
      'position': x.dataset.position,
      'salary': Number(x.dataset.salary.replace(/[^0-9.-]+/g, '')),
      'age': +x.dataset.age,
    };
  });
}

sortList(employees);
```

src/scripts/main.js
```diff
@@ -1,3 +1,43 @@
 'use strict';
 
 // write code here
+function sortList(ul) {
+  const employees = getEmployees(ul);
+
+  employees.sort((a, b) => b.salary - a.salary);
+
+  const formatter = new Intl.NumberFormat('en-US', {
```
Viktor-Kost, line 9:
```suggestion
function sortList(list) {
  const employees = getEmployees(list);

  [...employees].sort((a, b) => b.salary - a.salary);

  const formatter = new Intl.NumberFormat('en-US', {
```

Viktor-Kost (changes requested):

src/scripts/main.js
silcki: yes, that is great when you have only one UL on the page, but if not?

PR #820, opened 2023-04-04
1st version

Grechana:
https://grechana.github.io/js_task_parse_list_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,15 @@
 'use strict';
 
-// write code here
+const element = document.body.children[1].children;
+const sortlist = [...element].sort((x, y) =>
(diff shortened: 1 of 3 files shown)


polosanya (changes requested):
Almost done!

src/scripts/main.js
```diff
@@ -1,3 +1,21 @@
 'use strict';
 
-// write code here
+const peopleList = document.body.children[1].children;
```
polosanya, line 3: Better use querySelector (or querySelectorAll) instead to make your code cleaner
https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector

polosanya (changes requested):
A few more comments left)
Feel free to ask for help in chat if you have any difficulties)

qoa11a (approved):

PR #822, opened 2023-04-05
Develop

YuriiHlushenko:
https://YuriiHlushenko.github.io/js_task_parse_list_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,14 @@
 'use strict';
 
-// write code here
+const list = document.querySelectorAll('li');
+const sortedList = [...list].sort((a, b) =>
+  +b.attributes[1].value.replace(/[^\d]/g, '')
+  - +a.attributes[1].value.replace(/[^\d]/g, ''));
+
+const ul = document.querySelector('ul');
+
+ul.innerHTML = sortedList.map(li => `
+  <li>
+    ${li.innerText}
+  </li>
+`).join('');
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
...
+  sortedList.map((li, index) => employees.push(
+    {
+      name: li.innerText.trim(),
+      position: index + 1,
+      salary: li.attributes[1].value,
+      age: li.attributes[2].value,
+    }
+  ));
+
+  return employees;
```
IvanFesenko, line 32: You recive list as argument, why do you use list from upper scope?
use `list` instead `sortedList`

src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
...
+    +b.attributes[1].value.replace(/[^\d]/g, '')
+    - +a.attributes[1].value.replace(/[^\d]/g, ''));
+
+  const ul = document.querySelector('ul');
+
+  ul.innerHTML = sortedList.map(li => `
+    <li>
+      ${li.innerText}
+    </li>
+  `).join('');
```
IvanFesenko, line 17: Try do not use innerHTML. Use [append](https://developer.mozilla.org/en-US/docs/Web/API/Element/append) instead

IvanFesenko (changes requested):

mykhalenych (approved):

PR #826, opened 2023-04-08
solution

IamLoren:


src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
...
+  <h1>List of employees</h1>
+  <ul>
+    ${sortedList.map(item => `
+    <li 
+      data-name = '${item.textContent}'
+      data-position='${item.dataset.position}' 
+      data-salary='${item.dataset.salary}'
+      data-age= '${item.dataset.age}'>
+    ${item.textContent}
+    </li>`).join('')}
```
Volodymyr-Melnyk-Fron10der, line 26: consider creating li elements and appending it into the body

Volodymyr-Melnyk-Fron10der: document.createElement()

Volodymyr-Melnyk-Fron10der (changes requested):
plz be consistent with the task, I do not see any functions in your solution
image

src/scripts/main.js
```diff
@@ -1,3 +1,59 @@
 'use strict';
 
-// write code here
+const list = document.querySelectorAll('li');
+const ul = document.querySelector('ul');
+
+// сортування за з.п
```
OleziO, line 6: Remove all comments from the code.

OleziO (changes requested):

qoa11a (approved):

PR #827, opened 2023-04-08
solution

Pavlo-Petrashevskyi:


src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
 'use strict';
 
-// write code here
+const people = document.querySelectorAll('li');
+const list = document.querySelector('ul');
+
+function change(salary) {
```
Volodymyr-Melnyk-Fron10der, line 6: change what? make more meaningful name

src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+  return +salary.slice(1).replaceAll(',', '');
+};
+
+const sortList = listToSort => {
+  const currentList = Array.prototype.slice.call(listToSort, 0);
```
Volodymyr-Melnyk-Fron10der, line 11: what is the purpose of it? you wanted to make a copy or what?)  Because listToSort is definitely equals currentList  now. Consider to remove it or use JSON.parse/stringify to make a copy
```suggestion
  const currentList = Array.prototype.slice.call(listToSort, 0);
```

src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+
+  currentList
+    .sort((a, b) => change(b.dataset.salary) - change(a.dataset.salary))
+    .map(x => list.append(x));
```
Volodymyr-Melnyk-Fron10der, line 15: avoid name like a,b,x,i and other letters, create meaningful names. item, prevItem, curItem, employee, person, element and so on

src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+
+  return currentList;
+};
+
+function getEmployees(listToObjectArray) {
+  const currentList = Array.prototype.slice.call(listToObjectArray, 0);
```
Volodymyr-Melnyk-Fron10der, line 21: same

src/scripts/main.js
Volodymyr-Melnyk-Fron10der: map is useless here, you do not use a returned array, so just use forEach

src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+
+  for (let i = 0; i < currentList.length; i++) {
+    const newPerson = {
+      name: currentList[i].textContent
+        .replaceAll('\n', '')
+        .trim(),
+      position: currentList[i].dataset.position,
+      salary: currentList[i].dataset.salary,
+      age: currentList[i].dataset.age,
```
Volodymyr-Melnyk-Fron10der, line 30: you can spread dataset like that
```suggestion
      ...currentList[i].dataset,
```

src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+    .sort((a, b) => change(b.dataset.salary) - change(a.dataset.salary))
+    .map(x => list.append(x));
+
+  return currentList;
+};
+
+function getEmployees(listToObjectArray) {
+  const currentList = Array.prototype.slice.call(listToObjectArray, 0);
+
+  for (let i = 0; i < currentList.length; i++) {
```
Volodymyr-Melnyk-Fron10der, line 23: here you can use map, and return the array right away

Volodymyr-Melnyk-Fron10der (changes requested):

src/scripts/main.js
Pavlo-Petrashevskyi: I`ll agree, change won`t do.

Pavlo-Petrashevskyi: done

src/scripts/main.js
Pavlo-Petrashevskyi: I have tried to use method 'sort', but it dosn`t work with NoteList, I found this solution in the Internet? but I guess it works the same as [...ListToSort].

src/scripts/main.js
Pavlo-Petrashevskyi: Done

src/scripts/main.js
Pavlo-Petrashevskyi: Done

src/scripts/main.js
Pavlo-Petrashevskyi: Done

src/scripts/main.js
Pavlo-Petrashevskyi: Done

src/scripts/main.js
Pavlo-Petrashevskyi: Done

polosanya (approved):
Good job!

src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
 'use strict';
 
-// write code here
+const people = [...document.querySelectorAll('li')];
+const list = document.querySelector('ul');
+
+function transfStrToNum(salary) {
```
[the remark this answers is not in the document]
Pavlo-Petrashevskyi: Cool, thanks)

PR #842, opened 2023-04-20
done, but i don`t know what for getEmployees need

victor-buldenko:
[DEMO LINK](https://victor-buldenko.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,42 @@
 'use strict';
 
-// write code here
+function getNumberFromSalary(item) {
+  const result = +[...item.dataset.salary].map((char) => {
+    if (isNaN(char)) {
+      return;
+    }
+
+    return char;
+  }).join('');
+
+  return result;
+}
+
+function getEmployees(arr) {
+  const result = [];
+
+  arr.forEach((el) => {
+    result.push({
+      name: el.innerText,
+      position: el.dataset.position,
+      salary: el.dataset.salary,
+      age: el.dataset.age,
+    });
+  });
+
+  return result;
+}
+
+const list = document.querySelector('UL');
+const employees = [...document.getElementsByTagName('li')];
+
+employees.sort((a, b) => {
+  if (getNumberFromSalary(a) - getNumberFromSalary(b) > 0) {
+    return -1;
+  }
+});
+
+employees.forEach((el) => list.append(el));
+
+// eslint-disable-next-line no-unused-vars
+const employeesObjectsList = getEmployees(employees);


src/scripts/main.js
```diff
@@ -1,3 +1,42 @@
...
+  arr.forEach((el) => {
+    result.push({
+      name: el.innerText,
+      position: el.dataset.position,
+      salary: el.dataset.salary,
+      age: el.dataset.age,
+    });
+  });
+
+  return result;
```
IvanFesenko, line 27: use Array.map() instead forEach

src/scripts/main.js
```diff
@@ -1,3 +1,42 @@
...
+}
+
+const list = document.querySelector('UL');
+const employees = [...document.getElementsByTagName('li')];
+
+employees.sort((a, b) => {
+  if (getNumberFromSalary(a) - getNumberFromSalary(b) > 0) {
+    return -1;
+  }
```
IvanFesenko, line 36:
```suggestion
  return getNumberFromSalary(a) - getNumberFromSalary(b)
```

src/scripts/main.js
```diff
@@ -1,3 +1,42 @@
...
+});
+
+employees.forEach((el) => list.append(el));
+
+// eslint-disable-next-line no-unused-vars
+const employeesObjectsList = getEmployees(employees);
```
IvanFesenko, line 42:
```suggestion
getEmployees(employees);
```

IvanFesenko (changes requested):

src/scripts/main.js
victor-buldenko: i don`t want mutate array, 1 variable = 1 data.

src/scripts/main.js
victor-buldenko: if use your code, we must reverse array, i don`t want +1 unneeded step, so you need talk to me why your is best practice.

src/scripts/main.js
victor-buldenko: why forEach is bad in this task?

loralevitska: For the fact that you actually operate in memory with 2 arrays, one is iterated over and you push values into the other at each iteration of the cycle. The map method is not a mutating method, refresh your knowledge with this [link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map)

Please correct the code according to the recommendations
Also if you have any questions or if you didn't understand the mentor's comment, please ask them in fe_chat and tag mentors.

loralevitska (changes requested):
Almost done) Only one comment remains to be corrected 
Check unresolved conversation

src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+function getEmployees(arr) {
+  return arr.map((el) => {
+    return {
+      name: el.innerText,
+      position: el.dataset.position,
+      salary: el.dataset.salary,
+      age: el.dataset.age,
+    };
+  });
+}
```
ivbutsykin, line 24:
```suggestion
function getEmployees(arr) {
  return arr.map((el) => ({
      name: el.innerText,
      position: el.dataset.position,
      salary: el.dataset.salary,
      age: el.dataset.age,
    }));
}
```

ivbutsykin (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+  const result = +[...item.dataset.salary].map((char) => {
+    if (isNaN(char)) {
+      return;
+    }
+
+    return char;
+  }).join('');
+
+  return result;
+}
```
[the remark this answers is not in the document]
victor-buldenko: > This can be written in better way. For example, you can try to do sth like that:
> 
> ```
> function getNumberFromSalary(item) {
>     return parseInt(item.dataset.salary.replace(/[$,]/g, ''));
> }
> ```

Regexp not studied yet

PR #846, opened 2023-04-21
Parse List was created

KirillMaslov:


src/scripts/main.js
```diff
@@ -1,3 +1,52 @@
...
+    this.age = age;
+  }
+}
+
+function sortList() {
+  const employeesArr = [ ...list.children ];
+  let resultString = '';
+
+  employeesArr.sort((a, b) => {
+    return Number(b.dataset.salary.substring(1).replaceAll(',', ''))
```
SanyaBratashchuk, line 19: move `salary.substring(1).replaceAll(',', '')` this logic to the separate function

src/scripts/main.js
```diff
@@ -1,3 +1,52 @@
 'use strict';
 
-// write code here
+const list = document.body.querySelector('ul');
+
```
SanyaBratashchuk, line 4: const eployees = [...list.children]
```suggestion
const eployees = [...list.children]
```
To avoid double work on 15 and 41 lines

SanyaBratashchuk (changes requested):

ivanchuk-yurii (approved):

PR #847, opened 2023-04-22
solution v1

DanyloPasko:
[DEMO LINK](https://DanyloPasko.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,37 @@
 'use strict';
 
-// write code here
+const worker = document.querySelector('ul');
+const personList = [...worker.children];
+
+function sortSalary(list) {
+  const sortedEmployees = list.sort((a, b) => {
+    return (convertToNum(b.dataset.salary)
+      - convertToNum(a.dataset.salary));
+  });
+
+  for (const employee of sortedEmployees) {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
...
+  }
+}
+
+function getEmployees(list) {
+  const employees = list.map(elem => {
+    return {
+      name: elem.innerText,
+      position: elem.dataset.position,
+      salary: elem.dataset.salary,
+      age: elem.dataset.age,
```
Volodymyr-Melnyk-Fron10der, line 23: you can use spread here
```suggestion
      ...elem.dataset,
```

src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
...
+    };
+  });
+
+  return employees;
+}
+
+function convertToNum(str) {
+  const corectNumb = str.slice(1).split(',').join('');
+
+  return corectNumb;
```
Volodymyr-Melnyk-Fron10der, line 33: could be simplified
```suggestion
  return  str.slice(1).split(',').join('');
```

Volodymyr-Melnyk-Fron10der (approved):
well done, just check a few moments

PR #849, opened 2023-04-27
Parse-list

Soi4An:
DEMO - https://Soi4An.github.io/js_task_parse_list_DOM/
I will be happy to read comments about improvements. Don't fully understand what is mean "list property" in Steps to do this challenge, and why do we need the second function "getEmployees"... Thank for your attention

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,10 +1,9 @@
 'use strict';
 
 const list = document.querySelector('ul');
+const itemsList = [...document.querySelectorAll('li')];
 
-function sortListSalary(ul) {
-  const items = [...list.querySelectorAll('li')];
-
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,22 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('ul');
+
+function sortListSalary(ul) {
+  const items = [...list.querySelectorAll('li')];
```
IvanFesenko, line 6: Do not use querySelectorAll in this function. It should recieve this list as argument

IvanFesenko (changes requested):
getEmployees required for pratice

qoa11a (approved):
Good job! 👍

PR #853, opened 2023-04-28
solution

mariuszgit:


Radoslaw-Czerniawski (changes requested):
I get a 404 on the demo link

src/scripts/main.js
```diff
@@ -1,3 +1,51 @@
...
+function sortList(array) {
+  const sortedList = [...array.querySelectorAll('li')].sort(helperFn.sort);
+
+  // clear list
+  list.innerHTML = '';
+  // add items
+  sortedList.forEach(item => helperFn.appendToList(item));
+}
+
+function getEmployees(array) {
```
Radoslaw-Czerniawski, line 35: The parameter here implies an array of nodes, but you are passing a single one, rename it or grab all `li`'s at the beginning  of the file, later it may be refactored and shortened to:
```js
const getEmployees = (nodesList) =>
  [...nodesList].map((employee) => ({
    name: employee.innerText,
    position: employee.dataset.position,
    salary: employee.dataset.salary,
    age: employee.dataset.age,
  }));
```

src/scripts/main.js
```diff
@@ -1,3 +1,51 @@
+/* eslint-disable max-len */
 'use strict';
 
-// write code here
+const list = document.querySelector('ul');
```
Radoslaw-Czerniawski, line 4: In both functions you are querying all the nodes from this list, it would be more optimal to search the tree for them once here with querySelectorAll, and grab the ul element only in sortList

Radoslaw-Czerniawski: Also I would target the nodes not by the generic element `li` but rather by the specific data attribute like `[data-salary]`

src/scripts/main.js
```diff
@@ -1,3 +1,51 @@
...
+
+const helperFn = {
+  sort(prev, next) {
+    return helperFn.convertToNumber(next.dataset.salary) - helperFn.convertToNumber(prev.dataset.salary);
+  },
+  convertToNumber(number) {
+    return +number.replace('$', '').replace(',', '.');
+  },
+  appendToList(item) {
```
Radoslaw-Czerniawski, line 13: This method is redundant, there is no need to create copies of these items, you can just append their original nodes

src/scripts/main.js
```diff
@@ -1,3 +1,51 @@
...
+  },
+};
+
+function sortList(array) {
+  const sortedList = [...array.querySelectorAll('li')].sort(helperFn.sort);
+
+  // clear list
+  list.innerHTML = '';
+  // add items
+  sortedList.forEach(item => helperFn.appendToList(item));
```
Radoslaw-Czerniawski, line 32: If you just append those sorted nodes to the ul `list.innerHTML = '';` will be redundant
```js
sortedNodes.forEach(node => ul.append(node));
```

src/scripts/main.js
```diff
@@ -1,3 +1,51 @@
...
+        data-salary="${item.dataset.salary}"
+        data-age="${item.dataset.age}"
+      >
+      ${item.innerText}
+      </li>
+    `;
+  },
+};
+
+function sortList(array) {
```
Radoslaw-Czerniawski, line 26: The name of this parameter is incorrect as later you are passing here a single ul node rather than array of nodes. Rename it or adjust passed value.

Radoslaw-Czerniawski (changes requested):
Works very well, however needs some refactoring

Radoslaw-Czerniawski (approved):
Very nice, looks clean, left one small nit pick, wrote to you on chat about the strange innerText value mentioned during QnA

PR #857, opened 2023-05-02
Solution

Tarasova09:
- [DEMO LINK](https://Tarasova09.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,27 @@
 'use strict';
 
-// write code here
+const li = document.querySelectorAll('li');
+
+function sortList(list) {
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,27 @@
 'use strict';
 
-// write code here
+const li = document.querySelectorAll('li');
```
etojeDenys, line 3: Consider renaming the variable 'li' to something more descriptive like 'employeeList' for better readability.

etojeDenys (changes requested):
sorting does not work
image
image

polosanya (approved):
Nice!

PR #876, opened 2023-05-12
add parse_list_DOM solution

DidenkoDenis306:
- [DEMO LINK](https://DidenkoDenis306.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,29 @@
 'use strict';
 
-// write code here
+const listOfEmployees = document.querySelector('ul');
+const employees = [...listOfEmployees.children];
+
+function sortList(list) {
+  function employeeSalary(employee) {
+    return employee.dataset.salary
+      .slice(1)
+      .replace(/,/gi, '');
+  };
+
+  const sortedList = list
+    .sort((a, b) => +employeeSalary(b) - +employeeSalary(a));
+
+  listOfEmployees.append(...sortedList);
+};
+
+function getEmployees(list) {
+  return list.map(employee => ({
+    name: employee.innerText,
+    position: employee.dataset.position,
+    salary: employee.dataset.salary,
+    age: employee.dataset.age,
+  }));
+};
+
+sortList(employees);
+getEmployees(employees);
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -3,15 +3,15 @@
 const listOfEmployees = document.querySelector('ul');
 const employees = [...listOfEmployees.children];
 
-function sortList(list) {
-  function employeeSalary(employee) {
-    return employee.dataset.salary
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+
+function sortList(list) {
+  function employeeSalary(employee) {
+    return employee.dataset.salary
+      .slice(1)
+      .replace(/,/gi, '');
+  };
+
+  const sortedList = list
+    .sort((a, b) => +employeeSalary(b) - +employeeSalary(a));
```
33Orange, line 14: Consider using Number() instead of + for converting a string to a number.

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
-// write code here
+const listOfEmployees = document.querySelector('ul');
+const employees = [...listOfEmployees.children];
+
+function sortList(list) {
+  function employeeSalary(employee) {
+    return employee.dataset.salary
+      .slice(1)
+      .replace(/,/gi, '');
+  };
```
33Orange, line 11: Let's immediately return a number from this function instead of doing it later.
You can use Number() to convert string to number.

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
 'use strict';
 
-// write code here
+const listOfEmployees = document.querySelector('ul');
+const employees = [...listOfEmployees.children];
+
+function sortList(list) {
+  function employeeSalary(employee) {
+    return employee.dataset.salary
+      .slice(1)
```
33Orange, line 9: Why you use slice here?

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
 'use strict';
 
-// write code here
+const listOfEmployees = document.querySelector('ul');
+const employees = [...listOfEmployees.children];
+
+function sortList(list) {
+  function employeeSalary(employee) {
+    return employee.dataset.salary
```
33Orange, line 8: Remove this function outside of `sortList`. 
And transfer the value that you need to change there immediately, instead of getting it inside the function

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
 'use strict';
 
-// write code here
+const listOfEmployees = document.querySelector('ul');
+const employees = [...listOfEmployees.children];
+
+function sortList(list) {
+  function employeeSalary(employee) {
```
33Orange, line 7: Rename this function to something more descriptive

33Orange (changes requested):
Good job.

A few moments:
- Provide demo link
- Move function `employeeSalary` outside `sortList`. Also make so that this function would not depend on a job context. It should just take the string it needs to process and return a number.

PR #878, opened 2023-05-14
Solution

AV1788:
[DEMO LINK](https://AV1788.github.io/js_task_parse_list_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,54 @@
...
+  const transformTextPhase1 = text.split('');
+
+  transformTextPhase1.shift();
+
+  const transformTextPhase2 = transformTextPhase1.join('');
+  const transformTextPhase3 = transformTextPhase2.split(',').join('');
+  const result = Number(transformTextPhase3);
+
+  return result;
+}
```
anstsot, line 15: it can be simplified
```suggestion
function convNumber(text) {
  return Number(text.replace(/[^0-9.-]+/g, ''));
}
```

src/scripts/main.js
```diff
@@ -1,3 +1,54 @@
...
+
+function sortList(list) {
+  const items = list.querySelectorAll('li');
+  const itemsArray = [ ...items ];
+
+  for (const elem of itemsArray) {
+    const first = elem.dataset.salary;
+
+    elem.dataset.salary = convNumber(first);
+  }
```
anstsot, line 25: you don't need to overwrite the value. just use the function `convNumber` directly in `.sort` method below

src/scripts/main.js
```diff
@@ -1,3 +1,54 @@
...
+  const arr = [];
+
+  for (const elem of itemsArray) {
+    const obj = {};
+
+    obj.name = elem.innerText;
+    obj.position = elem.dataset.position;
+    obj.salary = elem.dataset.salary;
+    obj.age = elem.dataset.age;
+    arr.push(obj);
```
anstsot, line 47: you can directly push an object without creating `obj` variable:
```
arr.push({
  name: elem.innerText,
  ...
})
```

anstsot (changes requested):
It’s almost done! Let’s improve a little

TarasHoliuk (approved):

PR #893, opened 2023-05-22
done

Alessandranitti15:
[DEMO LINK](https://Alessandranitti15.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,32 @@
 'use strict';
 
-// write code here
+const employees = [...document.querySelectorAll('li')];
+const listUl = document.querySelector('ul');
+
+function convertSalary(salary) {
+  const noCurrency = salary.slice(1);
+
+  return +noCurrency.split(',').join('');
+}
+
+function sortList(salary) {
+  const sort = salary.sort((a, b) =>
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+
+  return +noCurrency.split(',').join('');
+}
+
+function sortList(salary) {
+  const sort = salary.sort((a, b) =>
+    convertSalary(b.dataset.salary) - convertSalary(a.dataset.salary));
+
+  return sort;
+}
```
anstsot, line 17: try to use more meaningful namings and there is no need to create additional variables
```suggestion
function sortList(list) {
  return [...list].sort((a, b) =>
    convertSalary(b.dataset.salary) - convertSalary(a.dataset.salary));
}
```

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+
+listUl.append(...sortList(employees));
```
anstsot, line 19: save sorted list to variable instead of calling it here and on line 31

anstsot (changes requested):
It’s almost done! Let’s improve a little

anstsot (approved):
Great job, approved!

PR #898, opened 2023-05-26
1st commit

ihor-halkin:
- [DEMO LINK](https://ihor-halkin.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,40 @@
 'use strict';
 
-// write code here
+const listOfEmploees = document.querySelectorAll('ul li');
+
+function sortList(list) {
+  const sortedList = [...list].sort((employee1, employee2) => {
+    const emp1 = toNumber(employee1.dataset.salary);
+    const emp2 = toNumber(employee2.dataset.salary);
+
+    return emp2 - emp1;
+  });
+
+  const employeeList = document.querySelector('ul');
+
+  sortedList.forEach((employee) => {
+    employeeList.appendChild(employee);
+  });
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+    const employeeName = employee.innerText;
+    const position = employee.dataset.position;
+    const employeesalary = employee.dataset.salary;
+    const age = employee.dataset.age;
+
+    return {
+      employeeName,
+      position,
+      employeesalary,
+      age,
```
agniya-i, line 35:
```suggestion
      name,
      position,
      salary,
      age,
```
Follow the schema from the task

src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
 'use strict';
 
-// write code here
+const listOfEmploees = document.querySelectorAll('ul li');
```
agniya-i, line 3:
```suggestion
const listOfEmployees = document.querySelectorAll('ul li');
```

src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+    return {
+      employeeName,
+      position,
+      employeesalary,
+      age,
+    };
+  });
+};
+
+getEmployees(sortList(listOfEmploees));
```
agniya-i, line 40: Looks like `sortList` function does not return anything, but `getEmployees` expect an array, does it works correctly?

agniya-i (changes requested):

PR #901, opened 2023-05-27
Added the solution

zynkd:
[DEMO LINK](https://zynkd.github.io/js_dom-task_parse_list/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,40 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('ul');
+
+function sortList(listItems) {
+  const sortedList = [...listItems].sort((a, b) => {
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+
+  return list;
+}
+
+function toNumber(listItem) {
+  return listItem.dataset.salary.slice(1).split(',').join('');
+}
+
+function getEmployees(listItems) {
+  return [...listItems].map((item, index) => ({
```
witflash, line 31: You can avoid using spreading array here. `map` method returns a new array always. So you can use just `return listItems.map(...)`

witflash (approved):
Well done 👍🏻

PR #920, opened 2023-06-06
solution

Vo7kov:
https://vo7kov.github.io/js_task_parse_list_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,45 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('ul');
+
+function sortList(roster) {
+  let rosterItems = roster.querySelectorAll('li');
+
+  [...rosterItems] = [...rosterItems].sort((a, b) => {
+    const numberA = convertToNumber(a.dataset.salary);
+    const numberB = convertToNumber(b.dataset.salary);
+
+    return numberB - numberA;
+  });
+
+  for (const item of [...rosterItems]) {
+    roster.append(item);
+  }
+}
+
+function convertToNumber(a) {
+  return +a.slice(1).split(',').join('');
+}
+
+function getEmployees(roster) {
+  const result = [];
+  const rosterItems = roster.querySelectorAll('li');
+
+  for (let index = 0; index < [...rosterItems].length; index++) {
+    const currentItem = [...rosterItems][index];
+
+    result.push(
+      {
+        name: currentItem.innerText,
+        position: index,
+        salary: currentItem.dataset.salary,
+        age: currentItem.dataset.age,
+      }
+    );
+  }
+
+  return result;
+}
+
+sortList(list);
+getEmployees(list);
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('ul');
+
+function sortList(roster) {
+  let rosterItems = roster.querySelectorAll('li');
```
Viktor-Kost, line 6: Consider using array spread syntax to convert the NodeList to an array, like `const rosterItems = [...roster.querySelectorAll('li')];`

src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
...
+
+  [...rosterItems] = [...rosterItems].sort((a, b) => {
+    const numberA = convertToNumber(a.dataset.salary);
+    const numberB = convertToNumber(b.dataset.salary);
+
+    return numberB - numberA;
+  });
+
+  for (const item of [...rosterItems]) {
```
Viktor-Kost, line 15: Replace the for loop with `forEach` for better readability: `[...rosterItems].forEach(item => roster.append(item));`

src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
...
+
+function convertToNumber(a) {
+  return +a.slice(1).split(',').join('');
+}
+
+function getEmployees(roster) {
+  const result = [];
+  const rosterItems = roster.querySelectorAll('li');
+
+  for (let index = 0; index < [...rosterItems].length; index++) {
```
Viktor-Kost, line 28: Replace the for loop with `forEach` for better readability, and remove the unnecessary array spread syntax: `rosterItems.forEach((currentItem, index) => {...});`

src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('ul');
+
+function sortList(roster) {
+  let rosterItems = roster.querySelectorAll('li');
+
+  [...rosterItems] = [...rosterItems].sort((a, b) => {
```
Viktor-Kost, line 8:
```suggestion
  rosterItems.sort((a, b) => {
```

src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
...
+function convertToNumber(a) {
+  return +a.slice(1).split(',').join('');
+}
+
+function getEmployees(roster) {
+  const result = [];
+  const rosterItems = roster.querySelectorAll('li');
+
+  for (let index = 0; index < [...rosterItems].length; index++) {
+    const currentItem = [...rosterItems][index];
```
Viktor-Kost, line 29: no needs to spread each time
```suggestion
    const currentItem = rosterItems[index];
```

Viktor-Kost (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,41 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('ul');
+
+function sortList(roster) {
+  const rosterItems = [...roster.querySelectorAll('li')];
```
VitaliyBondarenko1982, line 6: it can be
```suggestion
 const rosterItems = [...roster.querySelectorAll('li')];
 
function sortList(roster) {
```
looks like it can be find all li one time and use variable in both functions

VitaliyBondarenko1982 (approved):

PR #927, opened 2023-06-09
solution

dseq0:
- [DEMO LINK](https://dseq0.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,32 @@
 'use strict';
 
-// write code here
+const employeesList = document.querySelectorAll('li');
+
+function sortList(list) {
+  const salaryToFloat = (salary) => (
+    parseFloat(salary.replace('$', '').replace(',', '.'))
+  );
+
+  const sortedList = [...list].sort((a, b) => {
+    const salaryA = salaryToFloat(a.dataset.salary);
+    const salaryB = salaryToFloat(b.dataset.salary);
+
+    return salaryA === salaryB ? 0 : salaryA < salaryB ? 1 : -1;
+  });
+
+  document.querySelector('ul').append(...sortedList);
+}
+
+function getEmployees(list) {
+  return [...list].map(employee => {
+    return {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+function sortList(list) {
+  const salaryToFloat = (salary) => (
+    parseFloat(salary.replace('$', '').replace(',', '.'))
+  );
+
+  const sortedList = [...list].sort((a, b) => {
+    const salaryA = salaryToFloat(a.dataset.salary);
+    const salaryB = salaryToFloat(b.dataset.salary);
+
+    return salaryA === salaryB ? 0 : salaryA < salaryB ? 1 : -1;
```
OleziO, line 14: Replace with `return salaryA < salaryB ? 1 : -1;`

OleziO: Try to avoid nested ternary, it's bad practice.

OleziO (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+    const salaryB = salaryToFloat(b.dataset.salary);
+
+    return salaryA < salaryB ? 1 : -1;
+  });
+
+  document.querySelector('ul').append(...sortedList);
+}
+
+function getEmployees(list) {
+  return [...list].map(employee => {
```
VitaliyBodnar, line 21: I don't see any sense in destructuring [...list], because the method map() already return new array,
make sense to destructure employee

VitaliyBodnar (commented):
Well done 🔥
Let's improve your solution 🧑‍💻

src/scripts/main.js
VitaliyBodnar, line 21: I don't see any sense in destructuring [...list], because the method map() already return new array,
make sense to destructure employee

VitaliyBodnar (changes requested):

PR #931, opened 2023-07-18
Develop

Luk2asz:
- [DEMO LINK](https://Luk2asz.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -5,18 +5,6 @@ const list = document.querySelectorAll('li');
 const salaryToNumber = (string) => string.match(/\d+/g).join('');
 
 function sortList(elements) {
-  // const elementArray = Array.from(elements);
-
-  // elementArray.sort((a, b) => {
-  //   const x = salaryToNumber(b.getAttribute('data-salary'));
-  //   const y = salaryToNumber(a.getAttribute('data-salary'));
-
-  //   return x - y;
-  // });
-
-  // const ul = document.querySelector('ul');
-
-  // elementArray.forEach(x => ul.appendChild(x));
   const ul = document.querySelector('ul');
 
   const sortedEmployees = [...elements].sort(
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,42 @@
 'use strict';
 
-// write code here
+const list = document.querySelectorAll('li');
+
+function sortList(element) {
```
DorotaLeniecDev, line 5: Try to name parameters more descriptively
```suggestion
function sortList(elements) {
```

src/scripts/main.js
```diff
@@ -1,3 +1,42 @@
...
+  const elementArray = Array.from(element);
+
+  elementArray.sort((a, b) => {
+    let x = b.getAttribute('data-salary').substring(1);
+
+    x = Number(x.replace(',', ''));
```
DorotaLeniecDev, line 11: Try extracting helper function so your code is more readable:
```js
function salaryToNumber(string) {
  return +string.slice(1).replaceAll(',', '');
}
```

And you could use it here like:
```js
const employees = getEmployees(list);

employees.sort((a, b) => salaryToNumber(b.salary) - salaryToNumber(a.salary));
```

DorotaLeniecDev (changes requested):
Check my comments

src/scripts/main.js
```diff
@@ -1,3 +1,41 @@
...
+  }
+
+  return employeesList;
+}
+
+function salaryToNumber(string) {
+  return string.slice(1).replaceAll(',', '');
+}
+
+getEmployees(list);
```
Radoslaw-Czerniawski, line 41: This is OPTIONAL but we could [curry](https://javascript.info/currying-partials)  the results
1. first we need to make sortList return data that it's working on:
image

Radoslaw-Czerniawski: 2, now we can pass invocation of the sortList to getEmployees
image

Radoslaw-Czerniawski: Or do something even crazier, like implementing curry helper
image

Radoslaw-Czerniawski: then use it and pipe returned values from one function to another:
image

Radoslaw-Czerniawski (changes requested):
The last part is optional, however try implementing the previously suggested changes

DorotaLeniecDev (changes requested):
Please remove commented code

DorotaLeniecDev (changes requested):
RC

DorotaLeniecDev (approved):
Good job 👌

PR #932, opened 2023-07-21
js_task_parse_list_DOM

KarolinaWiktoria:
https://KarolinaWiktoria.github.io/js_task_parse_list_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,42 @@
 'use strict';
 
-// write code here
+function convertToNumber(salary) {
+  const cleanedString = salary.replace(/[$,]/g, '');
+  const number = +cleanedString;
+
+  return number;
+}
+
+const ul = document.querySelector('ul');
+const liArr = [...ul.querySelectorAll('li')];
+
+function sortList(propName) {
+  liArr.sort((a, b) => {
+    const salaryB = convertToNumber(b.dataset[propName]);
+    const salaryA = convertToNumber(a.dataset[propName]);
+
+    return salaryB - salaryA;
+  });
+
+  ul.innerHTML = '';
+  liArr.forEach(li => ul.append(li));
+}
+
+sortList('salary');
+
+function getEmployees() {
+  const result = [];
+
+  for (const person of liArr) {
+    result.push({
+      name: person.textContent.trim(),
+      position: person.dataset.position,
+      salary: person.dataset.salary,
+      age: person.dataset.age,
+    });
+  }
+
+  return result;
+}
+
+getEmployees();
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_task_parse_list_DOM/)
+    - [DEMO LINK](https://KarolinaWiktoria.github.io/js_task_parse_list_DOM/)
 2. Follow [this instructions](https://mate-academy.github.io/layout_task-guideline/)
     - Run `npm run test` command to test your code;
     - Run `npm run test:only -- -n` to run fast test ignoring linter;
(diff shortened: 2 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,42 @@
...
+  ul.innerHTML = '';
+  liArr.forEach(li => ul.append(li));
+}
+
+sortList('salary');
+
+function getEmployees() {
+  const result = [];
+
+  for (const person of liArr) {
```
darokrk, line 30: Avoid using a `for...of` loop to create a new array. Instead, use `map`.

```
const result = liArr.map(item => {
  return {
      name: item.textContent.trim(),
      salary: item.dataset.salary,
      age: item.dataset.age,
  }
})
```

src/scripts/main.js
```diff
@@ -1,3 +1,42 @@
 'use strict';
 
-// write code here
+function convertToNumber(salary) {
+  const cleanedString = salary.replace(/[$,]/g, '');
+  const number = +cleanedString;
```
darokrk, line 5: Please avoid using implicit coertion of types bcs it's hard to read, use rather `Number()` constructor function to convert it to number type

```suggestion
  const number = Number(cleanedString);
```

darokrk (changes requested):
Looks good 👍🏽  Few mentions to fixing from my site 🛠️

src/scripts/main.js
KarolinaWiktoria: oh sure :) thanks for your comments!

src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+    const salaryA = convertToNumber(a.dataset[propName]);
+
+    return salaryB - salaryA;
+  });
+
+  ul.innerHTML = '';
+  liArr.forEach(li => ul.append(li));
+}
+
+sortList('salary');
```
darokrk, line 25: this function call should be moved below and place right before `getEmployees();` call, to line 39

src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+  const number = Number(cleanedString);
+
+  return number;
+}
+
+const ul = document.querySelector('ul');
+const liArr = [...ul.querySelectorAll('li')];
+
+function sortList(propName) {
+  liArr.sort((a, b) => {
```
darokrk, line 14: Think this should also work:

```
function sortList(propName) {
  ul.innerHTML = '';
  liArr.sort((a, b) => {
    const salaryB = convertToNumber(b.dataset[propName]);
    const salaryA = convertToNumber(a.dataset[propName]);

    return salaryB - salaryA;
  }).forEach(li => ul.append(li));
}
```

src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
 'use strict';
 
-// write code here
+function convertToNumber(salary) {
```
darokrk, line 3: We can even also shorten this function to be: 

```
function convertToNumber(salary) {
  const number = Number(salary.replace(/[$,]/g, ''));
  return number;
}
```

src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+    return salaryB - salaryA;
+  });
+
+  ul.innerHTML = '';
+  liArr.forEach(li => ul.append(li));
+}
+
+sortList('salary');
+
+function getEmployees() {
```
darokrk, line 27: Shorten version:

```
function getEmployees() {
  const result = liArr.map(item => ({
      name: item.textContent.trim(),
      salary: item.dataset.salary,
      age: item.dataset.age,
    }));

  return result;
}
```

darokrk (approved):
Good job 💪🏽 ! Give approval but please look my new mention before move next, we can improve you code to be shorter 😃

src/scripts/main.js
KarolinaWiktoria: Ok nice suggestions. Thanks. I've fixed it once again.

darokrk (approved):

PR #935, opened 2023-08-01
sol

GGsmou:
[DEMO LINK](https://GGsmou.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,36 @@
 'use strict';
 
-// write code here
+const listElement = document.querySelector('ul');
+
+const parseSalary = (salary) => {
+  const parsedSalary = salary.replace('$', '').replace(',', '');
+
+  try {
+    return Number(parsedSalary);
+  } catch (error) {
+    return 0;
+  }
+};
+
+const sortList = (list) => {
+  const listElements = Array.from(list.children);
+
+  listElements.sort((a, b) =>
+    parseSalary(b.dataset.salary) - parseSalary(a.dataset.salary));
+
+  list.innerHTML = '';
+
+  listElements.forEach((element) => list.appendChild(element));
+};
+
+const getEmployees = (list) => {
+  return [...list.children].map((employee) => ({
+    name: employee.innerText,
+    position: employee.dataset.position,
+    salary: employee.dataset.salary,
+    age: employee.dataset.age,
+  }));
+};
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
...
+
+const parseSalary = (salary) => {
+  const parsedSalary = salary.replace('$', '').replace(',', '');
+
+  try {
+    return Number(parsedSalary);
+  } catch (error) {
+    return 0;
+  }
+};
```
OlEzhA0, line 13: why do we need this `try/catch` block? 
```suggestion
const parseSalary = (salary = '') => {
  const parsedSalary = salary.replace('$', '').replace(',', '');

  return +parsedSalary;
};
```

src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
...
+
+const sortList = (list) => {
+  const listElements = Array.from(list.children);
```
OlEzhA0, line 16: mb: 
```suggestion
  const listElements = [...list.children];
```

src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
...
+
+  listElements.sort((a, b) =>
+    parseSalary(b.dataset.salary) - parseSalary(a.dataset.salary));
+
+  list.innerHTML = '';
+
+  listElements.forEach((element) => list.appendChild(element));
```
OlEzhA0, line 23:
```suggestion
  listElements.forEach(list.appendChild);
```

OlEzhA0 (changes requested):

src/scripts/main.js
GGsmou: Unfortunately, it does not work as expected, it is possible to `.bind(list)` so your solution will work, but, for me, It makes solution harder to understand.

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
 'use strict';
 
-// write code here
+const listElement = document.querySelector('ul');
+
+const parseSalary = (salary) => {
+  const parsedSalary = salary.replace('$', '').replace(',', '');
+
+  return +parsedSalary;
```
DanilWeda, line 8: Alternative: Number(parsedSalary)

DanilWeda (approved):
Good job.

PR #973, opened 2023-12-02
Solution

YuraMisura:
[DEMO LINK](https://YuraMisura.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,35 @@
 'use strict';
 
-// write code here
+function sortList(list) {
+  const listEmployees = document.querySelector('ul');
+  const employees = document.querySelectorAll(`[${list}]`);
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
...
+
+  return sortedEmployees.forEach(person => {
+    listEmployees.appendChild(person);
+  });
+}
+
+function getEmployees(list) {
+  const employees = document.querySelectorAll(`${list}`);
+
+  return [...employees].map(person => {
```
VitaliyBondarenko1982, line 20:
```suggestion
  return employees.map(person => {
```
looks like redundant to make new array here. map returns new array and don't mutate current array

VitaliyBondarenko1982 (approved):
Good job.

PR #997, opened 2024-01-18
solution

vinogradova8:
[DEMO LINK](https://vinogradova8.github.io/js_task_parse_list_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,48 @@
...
+function convertToNumber(text) {
+  return +(text.replace('$', '').replace(',', ''));
+}
+
+const sortList = (input) => {
+  return [...input].sort((item1, item2) => {
+    if (convertToNumber(item1.dataset.salary)
+      === convertToNumber(item2.dataset.salary)) {
+      return item1.innerText.localeCompare(item2.innerText);
+    }
```
anastasiia-tilikina, line 16: i believe you don't need this check. just sort values as numbers

anastasiia-tilikina (changes requested):
one small comment

src/scripts/main.js
vinogradova8: but without this check the order of the employees does not meet the requirements of the task.

volodymyr-soltys97 (approved):
Good job 🔥

PR #1096, opened 2024-06-13
Add solution

AnastasiiaHombalevska:
🦊 [DEMO LINK](https://AnastasiiaHombalevska.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,8 +1,9 @@
 'use strict';
 
-const liElements = [...document.querySelectorAll('li')];
+const liElements = document.querySelectorAll('li');
+const listOfElements = document.querySelector('ul');
 
-const employes = liElements.map((li) => ({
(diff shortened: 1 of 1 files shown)


anastasiiavorobiova (changes requested):
Great start! Consider fixing the sort here

AnastasiiaHombalevska:
I fixed the sorting function and the display of salary in the final output.
The sorting test fails, although everything seems to be correct when outputting to the console.
I asked for help in the chat, but the result remained as it is now.
If possible, please take a look and tell me what could be the cause of the error.
Thank you in advance

DarkMistyRoom (changes requested):
You should rerender the list of employees after the sorting.
Better start a new thread in the chat - the communication will be faster

lerastarynets (approved):
lgtm 🚀

PR #1120, opened 2024-06-20
add solution

reznik-denis:
- [DEMO LINK](https://reznik-denis.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,44 @@
 'use strict';
 
-// write code here
+const list = document.querySelectorAll('li');
+const newSortedList = sortlist([...list]);
+const newlistEmployees = getEmployees(newSortedList);
+
+for (let i = 0; i < list.length; i++) {
+  list[i].innerHTML = newlistEmployees[i].name;
+  list[i].dataset.position = newlistEmployees[i].position;
+  list[i].dataset.salary = newlistEmployees[i].salary;
+  list[i].dataset.age = newlistEmployees[i].age;
+}
+
+function sortlist(sortedList) {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
...
+  listEmployees.map((item) => {
+    newlistEml.push({
+      name: item.innerText,
+      position: item.dataset.position,
+      salary: item.dataset.salary,
+      age: item.dataset.age,
+    });
+
+    return item;
+  });
```
anastasiia-tilikina, line 35: this is not good usage of map method
you can assign the result of map into variable since it returns a new array instead of pushing elements

anastasiia-tilikina (changes requested):
GJ

Let's just fix one moment here

src/scripts/main.js
reznik-denis: done

src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+
+function getEmployees(listEmployees) {
+  const newlistEml = listEmployees.map((item) => {
+    return {
+      name: item.innerText,
+      position: item.dataset.position,
+      salary: item.dataset.salary,
+      age: item.dataset.age,
+    };
+  });
```
etojeDenys, line 31:
```suggestion
  const newlistEml = listEmployees.map((item) => ({
      name: item.innerText,
      position: item.dataset.position,
      salary: item.dataset.salary,
      age: item.dataset.age,
    }));
```

etojeDenys (approved):

PR #1179, opened 2024-07-16
add task solution

Julia-na:
https://Julia-na.github.io/js_task_parse_list_DOM/


lerastarynets (approved):
You're making excellent progress!

PR #1205, opened 2024-07-25
add solution

RainerWerner44:
- [DEMO LINK](https://RainerWerner44.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,17 @@
 'use strict';
 
-// write code here
+function sortList(list) {
+  const itemsArray = Array.from(list.children);
+
+  itemsArray.sort(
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,17 @@
 'use strict';
 
-// write code here
+function sortList(list) {
+  const itemsArray = Array.from(list.children);
+
+  itemsArray.sort(
```
BudnikOleksii, line 6: - `itemsArray` is a bad name, a variable name should be descriptive
- You can use methods in a chain, instead of putting them from the new line

BudnikOleksii (approved):
👍

PR #1243, opened 2024-08-13
add solution

HokageHEL:
- [DEMO LINK](https://HokageHEL.github.io/js_task_parse_list_DOM/)


volodymyr-soltys97 (approved):
Good job!
The demo link does not need to be pasted as a block of code, it should be plain text so that the link can be clicked immediately
image

PR #1244, opened 2024-08-14
Solution

RostyslavSharuiev:
- [DEMO LINK](https://RostyslavSharuiev.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,25 @@
 'use strict';
 
-// write code here
+const employeesList = document.querySelector('ul');
+const employees = [...document.querySelectorAll('li')];
+
+const salaryToNum = (salary) => Number(salary.replace(',', '').slice(1));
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,25 @@
 'use strict';
 
-// write code here
+const employeesList = document.querySelector('ul');
+const employees = [...document.querySelectorAll('li')];
+
+const salaryToNum = (salary) => Number(salary.replace(',', '').slice(1));
```
VolodymyrKirichenko, line 6: We can use more clear name to this function.
For example:
`convertSalaryToNumber` or `parseSalary`.
Also we can use `parseInt` for this case(Up to u)

VolodymyrKirichenko (approved):
Good job!
I like this code 🔥

PR #1257, opened 2024-08-20
add task solution

StanislavKapytsia:
[DEMO LINK](https://StanislavKapytsia.github.io/js_task_parse_list_DOM/)
Sort list by salary in descending order. Get an array of employees. Write two functions:
first, which sorts the list by salary from data attributes
second, which returns an array of objects, where objects are employees.
The schema for the employee:
{
  name, 
  position,
  salary,
  age
}

result


etojeDenys (approved):
good job

PR #1284, opened 2024-09-09
task is done

BurchakDmitry:
[DEMO LINK](https://BurchakDmitry.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,25 @@
 'use strict';
 
-// write code here
+const items = [...document.querySelectorAll('[data-salary]')];
+const ul = document.querySelector('ul');
+
+function sortList(list) {
+  return list.sort(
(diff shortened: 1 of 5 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,25 @@
...
+const ul = document.querySelector('ul');
+
+function sortList(list) {
+  return list.sort(
+    (firstItem, secondItem) =>
+      getIterableItems(firstItem) - getIterableItems(secondItem),
+  );
+}
+
+function getIterableItems(item) {
```
TarasHoliuk, line 13: it works with and return one item/value
```suggestion
function getIterableItem(item) {
```
or better getSortableValue or getSalaryNumber, etc

TarasHoliuk (approved):
GJ 👍 
Approved with one comment (just FYI)

PR #1332, opened 2024-10-09
add task solution

AndriiZakharenko:
[DEMO LINK](https://AndriiZakharenko.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,14 @@
 'use strict';
 
-// write code here
+const salaryList = [...document.querySelectorAll('li[data-salary]')];
+
+salaryList.sort((a, b) => {
+  const item1 = parseInt(a.getAttribute('data-salary').replace('$', ''));
+  const item2 = parseInt(b.getAttribute('data-salary').replace('$', ''));
+
+  return item2 - item1;
+});
+
+salaryList.forEach((item) => {
+  document.querySelector('ul').appendChild(item);
+});
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,14 +1,24 @@
 'use strict';
 
-const salaryList = [...document.querySelectorAll('li[data-salary]')];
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,14 @@
 'use strict';
 
-// write code here
+const salaryList = [...document.querySelectorAll('li[data-salary]')];
+
+salaryList.sort((a, b) => {
+  const item1 = parseInt(a.getAttribute('data-salary').replace('$', ''));
+  const item2 = parseInt(b.getAttribute('data-salary').replace('$', ''));
```
igorkruz: You might  add error handling in case the salary attribute contains invalid values.
```suggestion
  const item1 = parseInt(a.getAttribute('data-salary').replace('$', '')) || 0;
  const item2 = parseInt(b.getAttribute('data-salary').replace('$', '')) || 0;
```

igorkruz (changes requested):
Your initial implementation is a great start, but it doesn't completely fulfill all the requirements of the task.

https://github.com/mate-academy/js_task_parse_list_DOM?tab=readme-ov-file#steps-to-do-this-challenge
image

igorkruz (changes requested):
Your solution covers the main points of the challenge, but there are a few improvements and requirements that need attention:

> Function Definitions: The requirement specifies that you should create a sortList function and a getEmployees function. Currently, the sorting logic is directly implemented without encapsulating it in a function. 

> Helper Function: You need to create a separate helper function to convert the salary string to a number. 

> getEmployees Function: This function is missing entirely. You need to define it to return the appropriate employee objects based on the list.

Review the task requirement list
https://github.com/mate-academy/js_task_parse_list_DOM?tab=readme-ov-file#steps-to-do-this-challenge

igorkruz (approved):
Well done 🔥

PR #1369, opened 2024-11-11
Task solution

Dominik-Dudek-code:
- [DEMO LINK](https://Dominik-Dudek-code.github.io/js_task_parse_list_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -8,8 +8,8 @@ const items = [...list.children];
 function sortList(elements) {
   [...elements]
     .sort((a, b) => {
-      const salaryA = formatedSalary(a.getAttribute('data-salary'));
-      const salaryB = formatedSalary(b.getAttribute('data-salary'));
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
...
+function getEmployees(elements) {
+  return elements.map((person) => ({
+    name: person.textContent.trim(),
+    position: person.getAttribute('data-position'),
+    salary: `$` + formattedSalary(person.getAttribute('data-salary')).toFixed(3),
+    age: Number(person.getAttribute('data-age')),
+  }));
+}
+
+getEmployees(items);
```
Zibi95, line 34: I was wondering what for is that, but I see that its in the requirements.

Zibi95 (approved):
Good job!

src/scripts/main.js
Dominik-Dudek-code: Yup, I also don't know why should i do it, without using it.. but well..

PR #1470, opened 2025-01-21
add task solution

Uusuff:


src/scripts/main.js
```diff
@@ -1,3 +1,38 @@
...
+for (let i = 0; i < array.length; i++) {
+  const user = array[i];
+
+  users.push({
+    name: user.innerText,
+    position: user.dataset.position,
+    salary: user.dataset.salary,
+    age: user.dataset.age,
+  });
+}
```
etojeDenys, line 16: you can simplify it using the map

etojeDenys (changes requested):

PR #1596, opened 2025-04-04
Develop

s1rserg:


s1rserg:
@mateacademy-ai-mentor 
there is no line in description of the task that i should store the value. linter gives me errors of unused variable

PR #1614, opened 2025-04-13
add task solution

TaniaP26:
https://TaniaP26.github.io/js_task_parse_list_DOM/


src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
...
+  const salary = +employ.dataset.salary.replace(/[^0-9.]/g, '');
+  const age = +employ.dataset.age;
+
+  employData.push({
+    element: employ,
+    position,
+    salary,
+    age,
+  });
+}
```
etojeDenys, line 18: should be in the getEmployees function, and try to simplify this logic using the .map method

etojeDenys (changes requested):

PR #1732, opened 2025-08-05
added task solution

horidor:
Link: [DEMO LINK](https://horidor.github.io/js_task_parse_list_DOM/)


horidor:
@mateacademy-ai-mentor The task includes only \<li> with text inside of them. I already target the most basic element, there cannot be other html elements inside of \<li> tags except for text.

horidor:
@mateacademy-ai-mentor The task specifies that i cannot change index.html file. There is no data-name attribute in index.html. Therefore, i can only access the text inside of it with element.innerText. That is why my code is written that way.

PR #1782, opened 2025-09-25
add task solution

NadiiaShishova:
- [DEMO LINK](https://NadiiaShishova.github.io/js_task_parse_list_DOM/)


src/main.js
```diff
@@ -0,0 +1,58 @@
+function parseSalary(salary) {
+  if (!salary) {
+    return 0;
+  }
+
+  return Number(salary.replace(/[^0-9]/g, '')) || 0;
```
FaiHamid: All characters except numbers are removed here.
It would be better to use the following way:
`return Number(salaryStr.replace(/[^0-9.-]+/g, ''));`

FaiHamid (changes requested):
Good job! 
Let`s just fix one detail.

etojeDenys (approved):
good job

PR #1791, opened 2025-10-04
solution

Maevalar:
[DEMO LINK](https://Maevalar.github.io/js_task_parse_list_DOM/)


Maevalar:
@mateacademy-ai-mentor I cannot take the name from a more specific element, since the username is contained in li, and according to the task, we cannot change the HTML markup. Lint also prohibits the use of console.log, so I cannot use it to display the result of the getEmployees function, but this does not interfere with the tests; all tests have been passed successfully.

